PR #2, opened 2023-01-20
Remove Twitter HackerOne reference.

tmcarr:
Problem

README still references the Twitter hacker one program.

Solution

Remove this reference. Without an obvious alternative, I chose to remove it entirely. It can be re-added once there is an alternative.


brayniac:
Nice catch. Thanks for opening this!

PR #6, opened 2023-01-31
add support for an http_header configuration option

yurivish:
This option is intended to allow [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) requests to the stats server.

I implemented an option for adding a single HTTP header rather than multiple because I don't think we need more than one right now and because I wasn't sure how to best support a "one or many" configuration style. I found [this](https://docs.rs/serde_with/latest/serde_with/struct.OneOrMany.html) but it looks like using it would add a dependency. Happy to make a change to that if needed.

There's also the matter of supporting preflight [OPTIONS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS) requests, but I think we can get away without needing to support them for now since the incoming requests will be [simple requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests) that don't trigger a preflight check.

I tested that the header is correctly added to a GET request with `curl`:

```
% curl -i 'http://0.0.0.0:4242/'
HTTP/1.1 200 OK
Server: tiny-http (Rust)
Date: Tue, 31 Jan 2023 18:26:31 GMT
Content-Type: text/plain; charset=UTF-8
Access-Control-Allow-Origin: *
Content-Length: 43

Welcome to rezolus
Version: 2.16.4-alpha.0
```

### DIFF
--- a/configs/example.toml
+++ b/configs/example.toml
@@ -33,7 +33,7 @@ listen = "0.0.0.0:4242"
 # reading_suffix = "count"
 
 # specify an HTTP header to be sent with each response from the stats server.
-# http_header = "Access-Control-Allow-Origin: *"
+# http_header = "Access-Control-Allow-Origin: <origin>"
 
 # Per-sampler configuration sections
 [samplers]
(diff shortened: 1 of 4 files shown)


brayniac (approved):
This looks good to me.

PR #9, opened 2023-04-06
feat: rewrite

brayniac:
Problem

Rezolus previously did not use BPF CO-RE for BPF samplers. This meant that it would compile the BPF programs at startup which results in long startup time and high CPU utilization at startup.

While addressing this it became apparent that a clean rewrite would be beneficial to simplify the code and make some different decisions in the overall design.

Solution

Rewrote the entire codebase.

Result

While not yet feature-complete with previous versions, we have greatly improve our BPF telemetry capabilities and reduced the runtime CPU utilization significantly.


thinkingfish:
.gitignore should include generated `bpf.rs` files under `src/samplers`

src/common/mod.rs
```diff
@@ -1,163 +1,169 @@
...
-    let line = line.trim();
-    let a: Vec<&str> = line.split('-').collect();
-    a.last()
-        .unwrap_or(&"0")
-        .parse::<u64>()
-        .map_err(|e| debug!("could not parse num cpus from file ({:?}): {}", path, e))
-        .map(|i| i + 1)
+#[macro_export]
+#[rustfmt::skip]
+macro_rules! counter {
```
thinkingfish, line 42: I'd like to see some documentations of `counter`, `heatmap`, `gauge`, and other combo types. There are also some considerations given to the range and granularity of the heatmap created, which would be good to document as well.

brayniac:
> .gitignore should include generated `bpf.rs` files under `src/samplers`

Moved the generated bpf.rs files into the `OUT_DIR` which is better than generating them into `src`.

src/common/bpf/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// The number of histogram buckets based on a rustcommon histogram with the
+/// parameters `m = 0`, `r = 8`, `n = 64`.
+///
+/// NOTE: this *must* remain in-sync across both C and Rust components of BPF
+/// code.
+const HISTOGRAM_BUCKETS: usize = 7424;
+const HISTOGRAM_PAGES: usize = 15;
+
+/// This function converts indices back to values for rustcommon histogram with
```
thinkingfish, line 27: How soon do you think we may want to introduce smaller histograms, say an order of magnitude or more smaller, either by virtue of much more limited value range, or because we can afford much coarser buckets?

src/samplers/hwinfo/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Serialize)]
+pub struct Hwinfo {
+    caches: Vec<Vec<Cache>>,
+    cpus: Vec<Cpu>,
+    memory: Memory,
+    network: Vec<Interface>,
+    nodes: Vec<Node>,
+}
+
+impl Hwinfo {
```
thinkingfish, line 33: My preference is to consolidate all the `Hwinfo` sources, e.g. `/proc/cpuinfo`, `proc/meminfo`, `/sys/devices/system/cpu/*` in this top level file. This way someone who is familiar with system info but not rezolus can quickly get an overall impression of what kind of info is being consumed by this module, and anybody can `cat` those paths to get examples of the data from a live system.

src/samplers/memory/proc_meminfo/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        let gauges = HashMap::from([
+            ("MemTotal:", &MEMORY_TOTAL),
+            ("MemFree:", &MEMORY_FREE),
+            ("MemAvailable:", &MEMORY_AVAILABLE),
+            ("Buffers:", &MEMORY_BUFFERS),
+            ("Cached:", &MEMORY_CACHED),
+        ]);
+
+        Self {
+            file: File::open("/proc/meminfo").expect("file not found"),
```
thinkingfish, line 38: Similar to `Hwinfo`, I recommend defining data source for each sampler as a constant at the module level.

src/common/bpf/mod.rs
brayniac: I don't see this as an urgent thing. I think consistency here gives us some nice guarantees around precision/error where we can document that it's less than some % and the users can be confident in that. It's only 60KB for the histogram in the BPF program.

The userspace heatmaps might be another story, but I think we can use lower precision or constrain the range there if appropriate.

src/samplers/hwinfo/mod.rs
brayniac: I agree. This particular sampler is not well-structured. Perhaps we should have some common reader functions and intermediate representations of the data coming from these files - there may be overlap with some of the files used for gathering telemetry too.

src/samplers/cpu/cpi/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+use perf_event::events::Hardware;
+use perf_event::{Builder, Group};
+
+#[distributed_slice(CPU_SAMPLERS)]
+fn init(config: &Config) -> Box<dyn Sampler> {
+    Box::new(Cpi::new(config))
+}
+
+pub struct Cpi {
+    prev: Instant,
```
thinkingfish, line 13: the use of `prev` `next` and the general ticking mechanism is used repeatedly, I think we may want to refactor that in provide a trait that's shared.

src/samplers/hwinfo/mod.rs
thinkingfish: good idea. Cutting an issue (#11) and we can skip it for now.

thinkingfish (approved):

PR #20, opened 2023-05-22
feat: add counts for some syscall groups

brayniac:
Adds counts for read, write, and poll related syscalls.


src/samplers/syscall/latency/mod.bpf.c
```diff
@@ -76,6 +116,22 @@ int sys_exit(struct trace_event_raw_sys_exit *args)
 		__sync_fetch_and_add(cnt, 1);
 	}
 
+	// for some syscalls, we track counts by "family" of syscall. check the
+	// lookup table and increment the appropriate counter
+	idx = 0;
+	if (syscall_id < 336) {
+		u32 counter_idx = syscall_lookup[syscall_id];
```
thinkingfish, line 123: We need a source for the 336 number and retrieval date, since it is likely to get updated in the future and also seems architecture dependent? I was looking at https://github.com/torvalds/linux/blob/master/arch/x86/entry/syscalls/syscall_64.tbl

brayniac: Yes, this will be arch dependent. I'm wondering if it's better to address that immediately, or merge and replace with a new strategy.

I'm considering having a map that the userspace writes into with the appropriate LUT for the runtime platform. This means that we can centralize the logic in userspace and keep the BPF a little more minimal. Also gives us some flexibility in terms of defining/enabling counts for groups by config if we want that in the future.

thinkingfish: I like this idea- put platform-specific initialization in one place before kicking off all the samplers.

 Do we have a global variable / switch for platform right now?

brayniac:
I believe the Rust cfg stuff gives us all we need for this, probably a good
time to start plumbing it through the codebase.

src/samplers/syscall/latency/mod.bpf.c
brayniac: I *think* the cfg stuff will get us most of the way there with target os and platform detection. I think it makes sense to start plumbing it in now.

src/samplers/syscall/latency/mod.bpf.c
```diff
@@ -76,6 +116,22 @@ int sys_exit(struct trace_event_raw_sys_exit *args)
...
+
+		if (counter_idx < 8) {
```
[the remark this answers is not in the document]
brayniac: Agree on moving the grouping numbers up. I think we still want to do a bounds check on the counter_idx after the lookup. This gives us a guarantee that the LUT data and counter grouping are in agreement and avoids accidentally writing into another CPU's counter group.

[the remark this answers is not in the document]
thinkingfish: Yes, I think a check is good to avoid accidental memory corruption. This defensive pattern is something I picked up while working on twemcache- we used to put 'NOT_REACHED()', which is a macro, on branches that should never be taken, and always complete the if/else statement to make it symmetrical

src/samplers/syscall/latency/mod.rs
```diff
@@ -49,7 +51,65 @@ impl Syscall {
...
+                .map_mut(&file)
+                .expect("failed to mmap() bpf syscall lut")
+        };
+
+        for (syscall_id, bytes) in syscall_lut.chunks_exact_mut(4).enumerate() {
+            let counter_offset = bytes.as_mut_ptr() as *mut u32;
+            if let Some(syscall_name) = syscall_numbers::native::sys_call_name(syscall_id as i64) {
+                let group = match syscall_name {
+                    // read related
+                    "read" | "pread64" | "readv" | "recvfrom" | "recvmsg" | "preadv"
```
thinkingfish, line 68: Shall we sort these alphabetically?

thinkingfish (approved):
This looks great now. One small comment but otherwise LGTM

PR #25, opened 2023-06-26
Perfevent2

yangxi:
Use per-CPU perf events to compute CPU metrics:

On each CPU, we set up a set of perf events in one perf group pinned on the CPU to count cycles, instructions, tsc, aperf, and mperf for computing the CPU metrics. 

cpu/perf_groups/active (gauge): the number of active per-CPU perf groups being measured.
cpu/cycles (counter): the number of retired cycles by all active CPUs.
cpu/instructions (counter): the number of retired instructions on all active CPUs.
cpu/ipkc (heatmap): the distribution of per-cpu instructions per thousand cycles in the sampling period.
cpu/ipkc/average (gauge): the average of IPKC on all CPUs.
cpu/ipus (heatmap): the distribution of per-cpu instructions per microsecond
cpu/ipus/average (gauge): the average of IPUS on all CPUs.
cpu/base_frequency/average (gauge): the average base CPU frequency
cpu/frequency (heatmap): the distribution of the running CPU frequency of each CPU in the sampling period.
cpu/frequency/average (gauge): the average running CPU frequency on all CPUs.


src/samplers/cpu/cpi/mod.rs
```diff
@@ -91,17 +179,86 @@ impl Sampler for Cpi {
...
+                        let cur_aperf = cur_group_data[cpu][&(self.groups[cpu].aperf)];
+                        let prev_aperf = prev_group_data[cpu][&(self.groups[cpu].aperf)];
+                        let ratio =
+                            (cur_aperf - prev_aperf) as f64 / (cur_mperf - prev_mperf) as f64;
+                        let running_frequency = (ratio * base_frequency as f64) as i64;
+                        avg_frequency += running_frequency;
+                        ipns += (cpu_ipc as f64 * ratio) as i64;
+                    }
+                    RUNNINGFREQUENCY.set(avg_frequency / nr_cpu);
+                    IPKC.set(ipc / nr_cpu);
```
yangxi, line 255: @brayniac Is this the right way to set the gauge value?

src/samplers/cpu/stats.rs
```diff
@@ -41,3 +41,15 @@ gauge!(
 
 counter!(CPU_CYCLES, "cpu/cycles");
 counter!(CPU_INSTRUCTIONS, "cpu/instructions");
+gauge!(IPKC, "cpu/ipkc", "Average Instructions Per Thousand Cycle");
```
yangxi, line 44: @brayniac  Does gauge support the float numbers? Otherwise I have to use IPKC is to get better precisions.

src/samplers/cpu/cpi/mod.rs
```diff
@@ -91,17 +179,86 @@ impl Sampler for Cpi {
...
+                        let prev_mperf = prev_group_data[cpu][&(self.groups[cpu].mperf)];
+                        let cur_aperf = cur_group_data[cpu][&(self.groups[cpu].aperf)];
+                        let prev_aperf = prev_group_data[cpu][&(self.groups[cpu].aperf)];
+                        let ratio =
+                            (cur_aperf - prev_aperf) as f64 / (cur_mperf - prev_mperf) as f64;
+                        let running_frequency = (ratio * base_frequency as f64) as i64;
+                        avg_frequency += running_frequency;
+                        ipns += (cpu_ipc as f64 * ratio) as i64;
+                    }
+                    RUNNINGFREQUENCY.set(avg_frequency / nr_cpu);
```
yangxi, line 254: @brayniac Using average runningfrequency, ipc, ipns probably is not a good idea especially when the CPU utilizations varies a lot cross CPUs. Maybe I should just use the histogram to track these three metrics before we have per-CPU metrics, just like what Rezolus does in the CPU frequency sampler?

src/samplers/cpu/cpi/mod.rs
```diff
@@ -13,71 +15,157 @@ fn init(config: &Config) -> Box<dyn Sampler> {
...
+    }
+}
+
+pub struct Cpi {
+    prev: Instant,
+    next: Instant,
+    interval: Duration,
+    groups: Vec<PerfGroup>,
+    // we need to keep the last perf event readings because we need to compute gauge IPC, IPNS, and running CPU frequency.
+    prev_group_data: Option<Vec<GroupData>>,
```
yangxi, line 129: @brayniac Is this the right way to track the previous perf counters so that it could compute the IPC for the sampling period?

src/samplers/cpu/cpi/mod.rs
```diff
@@ -91,17 +179,86 @@ impl Sampler for Cpi {
...
+                    let mut avg_frequency = 0;
+                    let mut ipns: i64 = 0;
+                    for cpu in 0..cur_group_data.len() {
+                        let cur_cycles = cur_group_data[cpu][&(self.groups[cpu].cycles)];
+                        let prev_cycles = prev_group_data[cpu][&(self.groups[cpu].cycles)];
+                        let cur_instructions =
+                            cur_group_data[cpu][&(self.groups[cpu].instructions)];
+                        let prev_instructions =
+                            prev_group_data[cpu][&(self.groups[cpu].instructions)];
+                        let cpu_ipc = (cur_instructions - prev_instructions) * 1000 / (cur_cycles - prev_cycles);
```
yangxi, line 242: @brayniac These are all computing per-CPU metrics, so once Rezolus supports per-CPU metrics, we can expose them easily.

src/samplers/cpu/cpi/mod.rs
```diff
@@ -91,17 +179,87 @@ impl Sampler for Cpi {
...
+                    }
+                    if counts.time_enabled() != counts.time_running() {}
+                    let cycles = counts[&(group.cycles)];
+                    let instructions = counts[&(group.instructions)];
+
+                    total_cycles += cycles;
+                    total_instructions += instructions;
+                    let tsc = counts[&(group.tsc)];
+                    base_frequency = tsc / running_us as u64;
+                    BASEFREQUENCY.set(base_frequency as i64);
```
yangxi, line 214: @brayniac Maybe we should move the basefrequency to the hwinfo() sampler since it's never changed.

src/samplers/cpu/cpi/mod.rs
brayniac: That seems reasonable to add histograms to track these.

src/samplers/cpu/cpi/mod.rs
brayniac: Yes. That will overwrite the current gauge reading with the provided value.

src/samplers/cpu/stats.rs
brayniac: Gauge is only i64. So it would be better to use some scaled variant like you're doing with IPKC.

src/samplers/cpu/cpi/mod.rs
brayniac: That's a good idea. We should probably calculate it once and store it somewhere. It's probably okay to refactor this later.

src/samplers/cpu/cpi/mod.rs
```diff
@@ -91,17 +179,87 @@ impl Sampler for Cpi {
...
+                .set(now, elapsed, total_instructions);
+            // update the IPC and IPNS
+            if let Some(prev_group_data) = &self.prev_group_data {
+                let nr_cpu = cur_group_data.len() as i64;
+                if cur_group_data.len() == prev_group_data.len() {
+                    let mut ipc: i64 = 0;
+                    let mut avg_frequency = 0;
+                    let mut ipns: i64 = 0;
+                    for cpu in 0..cur_group_data.len() {
+                        let cur_cycles = cur_group_data[cpu][&(self.groups[cpu].cycles)];
```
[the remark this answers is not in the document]
brayniac: I'll ignore it for now and wait for an update.

src/samplers/cpu/stats.rs
```diff
@@ -41,3 +41,19 @@ gauge!(
...
```
brayniac, line 44: Since these are globals, let's be sure to prefix with `CPU_`

src/samplers/cpu/stats.rs
```diff
@@ -41,3 +41,19 @@ gauge!(
...
+gauge!(
+    IPKNS,
+    "cpu/ipkns",
+    "Average Instructions Per Thousand Nanosecond"
+);
+gauge!(
+    BASEFREQUENCY,
```
brayniac, line 51: `CPU_BASE_FREQUENCY` would be a better name

src/samplers/cpu/stats.rs
```diff
@@ -41,3 +41,19 @@ gauge!(
...
+    "cpu/base_frequency",
+    "The base CPU frequency"
+);
+gauge!(
+    RUNNINGFREQUENCY,
```
brayniac, line 56: `CPU_RUNNING_FREQUENCY`

src/samplers/cpu/stats.rs
```diff
@@ -41,3 +41,19 @@ gauge!(
 
 counter!(CPU_CYCLES, "cpu/cycles");
 counter!(CPU_INSTRUCTIONS, "cpu/instructions");
+gauge!(IPKC, "cpu/ipkc", "Average Instructions Per Thousand Cycle");
+gauge!(
+    IPKNS,
```
brayniac, line 46: Is this a conventional way of expressing this? Why is it not IPUS (Instructions per Microsecond)? Is the scaling here for the same reason, to avoid loss of precision when representing as a i64 instead of a float? What's the benefit of exporting this separately from `cpu/instructions` and doing a `rate()` in the obs stack?

src/samplers/cpu/cpi/mod.rs
brayniac: Actually, in-context of the whole sampler, I'm not sure why you're using an Option<> around this when you populate the prev_group_data in the constructor.

src/samplers/cpu/cpi/mod.rs
```diff
@@ -91,17 +179,87 @@ impl Sampler for Cpi {
 
         let elapsed = (now - self.prev).as_secs_f64();
 
-        match self.group.read() {
-            Ok(counts) => {
-                for (metric, pc) in &mut self.counters {
-                    metric.set(now, elapsed, counts[&pc]);
+        let mut total_cycles = 0;
```
brayniac, line 182: This isn't very idiomatic. I'd prefer these variables are declared where they're defined. I'd expect you get a bunch of clippy errors for this...

src/samplers/cpu/cpi/mod.rs
```diff
@@ -91,17 +179,87 @@ impl Sampler for Cpi {
...
+        let mut total_instructions = 0;
+        let mut base_frequency = 1;
+        let mut cur_group_data = vec![];
+        for group in &mut self.groups {
+            match group.read_group() {
+                Ok(counts) => {
+                    let running_us;
```
brayniac, line 189: try `let running_us = match ... { ... };` instead. It's more idiomatic

src/samplers/cpu/cpi/mod.rs
```diff
@@ -91,17 +179,87 @@ impl Sampler for Cpi {
...
+                        }
+                        (_, _) => {
+                            error!(
+                                "The perf group {} has no enabled time and running time",
+                                group.cpu
+                            );
+                            continue;
+                        }
+                    }
+                    if counts.time_enabled() != counts.time_running() {}
```
brayniac, line 206: What's the point of this?

src/samplers/cpu/cpi/mod.rs
yangxi: Thanks, will do.

PR #33, opened 2023-07-25
feat: enable bpf by default

brayniac:
Makes the bpf dependencies linux specific and enables the bpf feature by default.

Change all bpf feature gates within the code to include the target_os to prevent issues when building on non-linux systems with the default feature set

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,6 @@ clap = "4.1.4"
 clocksource = "0.6.0"
 lazy_static = "1.4.0"
 libc = "0.2.141"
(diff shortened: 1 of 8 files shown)


.github/workflows/cargo.yml
```diff
@@ -94,7 +97,7 @@ jobs:
       matrix:
         os: [ ubuntu-latest, ubuntu-20.04 ]
         profile: [ release, debug ]
-        features: [ default, bpf ]
+        features: [ default, no-default ]
```
thinkingfish, line 100: `nondefault` ?

PR #36, opened 2023-07-26
feat: add tcp packet latency sampler

brayniac:
Adds a TCP packet latency sampler that measures the latency from as packet being received to being handled by userspace application.

### DIFF
--- a/src/samplers/tcp/packet_latency/mod.bpf.c
+++ b/src/samplers/tcp/packet_latency/mod.bpf.c
@@ -1,0 +1,128 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2023 Wenbo Zhang
+
+// NOTICE: this file is based off `tcppktlat.bpf.c` from the BCC project
+// <https://github.com/iovisor/bcc/> and has been modified for use within
+// Rezolus.
+
+// This BPF program probes TCP receive path gather statistics about the latency
+// from a packet being received to it being processed by the userspace
+// application.
+
+#include "../../../common/bpf/vmlinux.h"
+#include "../../../common/bpf/histogram.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+
+#define MAX_ENTRIES	10240
+#define AF_INET		2
+
+const volatile pid_t targ_pid = 0;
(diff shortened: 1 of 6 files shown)


src/samplers/tcp/packet_latency/mod.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+
+	if (targ_sport && targ_sport != BPF_CORE_READ(inet, inet_sport))
+		return 0;
+	if (targ_dport && targ_dport != BPF_CORE_READ(sk, __sk_common.skc_dport))
+		return 0;
+	th = (const struct tcphdr*)BPF_CORE_READ(skb, data);
+	doff = BPF_CORE_READ_BITFIELD_PROBED(th, doff);
+	len = BPF_CORE_READ(skb, len);
+	/* `doff * 4` means `__tcp_hdrlen` */
+	if (len <= doff * 4)
```
thinkingfish, line 62: let's wrap this in `{}` just for good measure.

src/samplers/tcp/packet_latency/mod.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+	u64 sock_ident = get_sock_ident(sk);
+	u64 id = bpf_get_current_pid_tgid(), *tsp;
+	u32 idx;
+	u64 now, delta_ns, *cnt;
+	u32 pid = id >> 32, tid = id;
+	struct event *eventp;
+	u16 family;
+
+	tsp = bpf_map_lookup_elem(&start, &sock_ident);
+	if (!tsp)
```
thinkingfish, line 82: ditto

src/samplers/tcp/packet_latency/mod.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+		return 0;
+
+	now = bpf_ktime_get_ns();
+
+	if (*tsp > now)
```
thinkingfish, line 87: ditto

thinkingfish (commented):
What happens when multiple packets are received before a user space copy is called? Would the timestamp in the map be overridden each time?

brayniac:
> What happens when multiple packets are received before a user space copy is called? Would the `start` timestamp in the map be overridden each time?

Yes, that appears to be correct. Should probably change the behavior there.

PR #60, opened 2023-08-23
Add support for cross compiling and arm64 debian package

swlynch99:
This PR fixes up the debian files and the build script to properly support cross compilation.


brayniac (commented):
LGTM

PR #63, opened 2023-08-28
feat: allow enabling/disabling of samplers

brayniac:
Adds configuration options that allow for enabling or disabling of samplers.


config.toml
```diff
@@ -1,4 +1,72 @@
...
+[samplers.block_io_latency]
+enabled = true
+
+# Produces various nVIDIA specific GPU metrics using NVML
+[samplers.gpu_nvidia]
+enabled = true
+
+# Memory utilization from /proc/meminfo
+[samplers.memory_meminfo]
+enabled = true
```
mihirn, line 35: Does this fail gracefully on machines without nvidia gpus? Or should be set it to disabled by default?

src/config/mod.rs
```diff
@@ -67,13 +105,127 @@ impl General {
...
+pub fn interval() -> String {
+    "10ms".into()
+}
+
+pub fn distribution_interval() -> String {
+    "50ms".into()
+}
+
+#[derive(Deserialize)]
+pub struct SamplerDefaults {
```
mihirn, line 121: Does this need to be its own structure? Couldn't `Config.defaults` be a `SamplerConfig`?

src/samplers/tcp/snmp/mod.rs
```diff
@@ -10,11 +10,15 @@ fn init(config: &Config) -> Box<dyn Sampler> {
     // if we have bpf enabled, we don't need to run this sampler at all
     if config.bpf() {
         Box::new(Nop::new(config))
+    } else if let Ok(s) = Snmp::new(config) {
+        Box::new(s)
     } else {
-        Box::new(Snmp::new(config))
+        Box::new(Nop::new(config))
```
mihirn, line 16: Curious as to what the difference between `Box::new(Nop {})` and this line is.

src/config/mod.rs
```diff
@@ -67,13 +105,127 @@ impl General {
     }
 }
 
+pub fn enabled() -> bool {
```
mihirn, line 108: Can't find any consumers of these three -- leftover from earlier?

config.toml
brayniac: Yes, this fails gracefully - an error is emitted during sampler init, but the sampler is effectively disabled at that point. It does not become part of the set of samplers and consumes no CPU time.

src/config/mod.rs
brayniac: These are used for the serde defaults which need to be functions.

mihirn (approved):

PR #67, opened 2023-09-27
update metriken to replace heatmaps

brayniac:
Updates metriken library to replace heatmaps with histograms which have better runtime performance characteristics and lower memory footprint.


src/common/bpf/distribution.rs
```diff
@@ -22,51 +23,62 @@ use super::*;
 pub struct Distribution<'a> {
     _map: &'a libbpf_rs::Map,
     mmap: memmap2::MmapMut,
-    prev: [u64; HISTOGRAM_BUCKETS],
-    heatmap: &'static Heatmap,
+    buffer: [u64; HISTOGRAM_BUCKETS],
```
mihirn, line 26: Shall we move this allocation to inside the else? Obviously it would suck it we are hitting that codepath, but since we expect never to, would end up saving some memory.

src/common/bpf/counters.rs
```diff
@@ -50,32 +51,49 @@ impl<'a> Counters<'a> {
         }
     }
 
-    pub fn refresh(&mut self, now: Instant, elapsed: f64) {
+    pub fn refresh(&mut self, elapsed: f64) {
         for value in self.values.iter_mut() {
```
mihirn, line 55: No strong opinion, but since this is perfomance-sensitive: would a `slice.fill` be faster here?

src/common/bpf/distribution.rs
brayniac: This line doesn't allocate? The allocation happens at L#44

mihirn: Sorry, that was imprecise. I was suggesting moving this out of the structure altogether and onto L#60 since, afaict, the buffer is only used in the misaligned case.

src/common/mod.rs
```diff
@@ -19,13 +17,13 @@ type Instant = clocksource::Instant<clocksource::Nanoseconds<u64>>;
 pub struct Counter {
     previous: Option<u64>,
     counter: &'static LazyCounter,
-    heatmap: Option<&'static Heatmap>,
+    heatmap: Option<&'static AtomicHistogram>,
```
mihirn, line 20: Rename heatmap to histogram?

src/common/mod.rs
```diff
@@ -19,13 +17,13 @@ type Instant = clocksource::Instant<clocksource::Nanoseconds<u64>>;
...
 }
 
 impl Counter {
     /// Construct a new counter that wraps a `metriken` counter and optionally a
     /// `metriken` heatmap.
-    pub fn new(counter: &'static LazyCounter, heatmap: Option<&'static Heatmap>) -> Self {
+    pub fn new(counter: &'static LazyCounter, heatmap: Option<&'static AtomicHistogram>) -> Self {
```
mihirn, line 26: Same here as well as in the comment.

src/common/mod.rs
```diff
@@ -112,15 +110,41 @@ macro_rules! heatmap {
...
+/// The heatmap configuration used here can record counts for all 64bit integer
+/// values with a maximum error of 0.78%. The heatmap covers a moving window of
+/// one minute with one second resolution.
+macro_rules! bpfhistogram {
+    ($ident:ident, $name:tt) => {
+        #[metriken::metric(
+            name = $name,
+            crate = metriken
+        )]
+        pub static $ident: metriken::RwLockHistogram = metriken::RwLockHistogram::new(7, 64);
```
mihirn, line 139: No strong opinion, just as a thought: is it worth making 7 and 64 consts so that it is easier to change them if we decide we want to track at a different granularity (7 being more likely to change).

src/common/mod.rs
```diff
@@ -112,15 +110,41 @@ macro_rules! heatmap {
             name = $name,
             crate = metriken
         )]
-        pub static $ident: metriken::Heatmap = metriken::Heatmap::new(0, 8, 64, core::time::Duration::from_secs(1), core::time::Duration::from_millis(100));
+        pub static $ident: metriken::AtomicHistogram = metriken::AtomicHistogram::new(7, 64);
```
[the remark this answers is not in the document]
brayniac: All heatmaps renamed to histograms

src/common/bpf/distribution.rs
```diff
@@ -16,57 +17,70 @@ use super::*;
...
-            *prev = val;
+        if buckets.len() == HISTOGRAM_BUCKETS {
+            let _ = self.histogram.update_from(&buckets);
+        } else {
+            warn!("mmap region misaligned or did not have expected number of values");
 
-            if delta > 0 {
-                let value = key_to_value(idx as u64);
-                let _ = self.heatmap.add(now, value as _, delta as _);
+            self.buffer.resize(HISTOGRAM_BUCKETS, 0);
```
mihirn, line 60: Oh, nice, this is much neater than I had in mind.

src/common/mod.rs
brayniac: This is not really possible to change without other code changes, the indexing in the BPF progs and in Rezolus need to agree exactly. We don't have a mechanism to replace the indexing function in the BPF prog if we change the histogram params.

mihirn: Ack, makes sense, thanks.

src/main.rs
```diff
@@ -3,10 +3,105 @@ use clap::{Arg, Command};
...
 use ringlog::*;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::SystemTime;
+use tokio::sync::RwLock;
 
 type Duration = clocksource::Duration<clocksource::Nanoseconds<u64>>;
 type Instant = clocksource::Instant<clocksource::Nanoseconds<u64>>;
 
+type HistogramSnapshots = HashMap<String, metriken::histogram::Snapshot>;
```
mihirn, line 14: Not for this changeset, but if this is the abstraction that both rpc-perf and rezolus end up using and it works well, it might be worth hoisting it into metriken.

src/main.rs
```diff
@@ -3,10 +3,105 @@ use clap::{Arg, Command};
...
+        for metric in metriken::metrics().iter() {
+            let any = if let Some(any) = metric.as_any() {
+                any
+            } else {
+                continue;
+            };
+
+            let key = metric.name().to_string();
+
+            if let Some(histogram) = any.downcast_ref::<metriken::AtomicHistogram>() {
```
mihirn, line 46: Not sure if the type system will allow it, but since both Atomic and RwLockHistogram return the same kind of snapshot, can the two arms be combined with an ||?

brayniac: Not unless we refactor to do something with traits. I can refactor this callsite slightly, but it's not a huge improvement.

brayniac: Made a small change here.

mihirn: Don't have much of an opinion on which pattern is better, but we may as well use the same one in `new` and `update`.

mihirn (approved):

PR #68, opened 2023-10-03
Extract the hwinfo sampler into a separate systeminfo crate

swlynch99:
I would like to use the hwinfo sampler (and whichever system info samplers get added in the future) from within systemslab. This PR moves the hwinfo sampler to a separate `systeminfo` crate and then makes rezolus depend on it.


swlynch99:
The PR seems to be failing cargo audit but that is not related to any changes I made within the PR.

mihirn (approved):

thinkingfish (approved):

brayniac (approved):

PR #87, opened 2023-10-20
feat: add TCP connection state sampler

brayniac:
Tracks TCP connections by state.

### DIFF
--- a/src/samplers/tcp/connection_state/mod.rs
+++ b/src/samplers/tcp/connection_state/mod.rs
@@ -1,0 +1,133 @@
+use crate::common::Nop;
+use crate::samplers::tcp::stats::*;
+use crate::samplers::tcp::*;
+use metriken::Gauge;
+use std::fs::File;
+use std::io::Read;
+use std::io::Seek;
+
+#[distributed_slice(TCP_SAMPLERS)]
+fn init(config: &Config) -> Box<dyn Sampler> {
+    if let Ok(s) = ConnectionState::new(config) {
+        Box::new(s)
+    } else {
+        Box::new(Nop::new(config))
+    }
+}
+
+const NAME: &str = "tcp_connection_state";
+
+pub struct ConnectionState {
+    prev: Instant,
+    next: Instant,
+    interval: Duration,
+    files: Vec<File>,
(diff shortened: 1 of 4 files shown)


src/samplers/tcp/stats.rs
```diff
@@ -30,6 +35,114 @@ counter_with_histogram!(
     "tcp/transmit/retransmit",
     "number of TCP segments retransmitted"
 );
+counter_with_histogram!(
+    TCP_CONN_ACTIVE_OPEN,
+    TCP_CONN_ACTIVE_OPEN_HISTOGRAM,
+    "tcp/connection/active_open",
+    "number of TCP active opens"
+);
+counter_with_histogram!(
```
mihirn, line 44: Where are these two histograms being used?

brayniac: Whoops. Wasn't supposed to include those in this PR.

src/samplers/tcp/connection_state/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+                    return;
+                }
+
+                for line in data.lines() {
+                    let parts: Vec<&str> = line.split_whitespace().collect();
+
+                    // find and increment the temporary gauge for this state
+                    if let Some(Ok(state)) = parts.get(3).map(|v| u8::from_str_radix(v, 16)) {
+                        if let Some((_, gauge)) = self.gauges.get_mut(state as usize - 1) {
+                            *gauge += 1;
```
mihirn, line 107: Not familiar with the data format so just checking: are we expecting to increment the gauge by only 1 every time as opposed to some value that is ready from the file?

mihirn (approved):

src/samplers/tcp/connection_state/mod.rs
brayniac: Yes, each open TCP socket is represented by a single line in the file. So we are grabbing the state field (4th column) and incrementing the corresponding state for each socket.

mihirn: Ack, thanks!

PR #89, opened 2023-10-24
adjust tcp connection state sampler interval in example config

brayniac:
Adjusts the TCP connection state sampler to use a 1s interval instead of the global default. This reduces the CPU utilization of Rezolus when running with the example config.


mihirn:
Worth putting approximately how much overhead this is causing in either the commit message or the comment? I realise this is hardware specific, but just as a ballpark estimate.

mihirn (approved):

PR #90, opened 2023-10-24
feat: add sampler for Rezolus itself

brayniac:
Adds a sampler to gather basic resource utilization metrics for Rezolus itself.

### DIFF
--- a/src/samplers/rezolus/rusage/mod.rs
+++ b/src/samplers/rezolus/rusage/mod.rs
@@ -1,0 +1,114 @@
+use super::stats::*;
+use super::*;
+use crate::common::Nop;
+
+const S: u64 = 1_000_000_000;
+const US: u64 = 1_000;
+const KB: i64 = 1024;
+
+#[distributed_slice(REZOLUS_SAMPLERS)]
+fn init(config: &Config) -> Box<dyn Sampler> {
+    if let Ok(rusage) = Rusage::new(config) {
+        Box::new(rusage)
+    } else {
+        Box::new(Nop {})
+    }
+}
+
+const NAME: &str = "rezolus_rusage";
+
+pub struct Rusage {
+    prev: Instant,
+    next: Instant,
+    interval: Duration,
+}
+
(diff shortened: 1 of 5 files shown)


src/samplers/rezolus/stats.rs
```diff
@@ -0,0 +1,74 @@
...
+use metriken::Gauge;
+use metriken::LazyCounter;
+
+#[metric(
+    name = "rezolus/cpu/usage/user",
+    description = "The amount of CPU time Rezolus was executing in user mode"
+)]
+pub static RU_UTIME: LazyCounter = LazyCounter::new(Counter::default);
+
+histogram!(
```
mihirn, line 13: Curious: is defining both of them separately the same as `counter_with_histogram!`?

src/samplers/rezolus/rusage/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+        if now < self.next {
+            return;
+        }
+
+        let elapsed = (now - self.prev).as_secs_f64();
+
+        self.sample_rusage(elapsed);
+
+        // determine when to sample next
+        let next = self.next + self.interval;
```
mihirn, line 61: Any reason not to just do `now + self.interval` in all cases? Seems like this would allow you to run a little bit over consistently until you are just about one interval behind and then do that anyway.

mihirn (approved):

src/samplers/rezolus/stats.rs
brayniac: It is. I switched over to the convenience macros for all these stats. If we start using labels, we'll need to go with the previous approach to add that metadata.

src/samplers/rezolus/rusage/mod.rs
brayniac: That might be reasonable. Deferring that as it applies to all samplers

mihirn: Ack, works for me.

PR #91, opened 2023-10-24
feat(sampler): change CPU features from a string to an array

swlynch99:
Currently, systeminfo stores the CPU features as a giant string. Semantically, though, it's an array of strings and it would be more useful to have the data representation reflect that.

### DIFF
--- a/crates/systeminfo/src/hwinfo/cpu.rs
+++ b/crates/systeminfo/src/hwinfo/cpu.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::collections::HashSet;
 use std::fs::File;
 use std::io::BufRead;
 use std::io::BufReader;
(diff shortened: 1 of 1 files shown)


thinkingfish (approved):

crates/systeminfo/src/hwinfo/cpu.rs
```diff
@@ -26,7 +26,7 @@ pub struct Cpu {
     pub microcode: Option<String>,
     pub vendor: Option<String>,
     pub model_name: Option<String>,
-    pub features: Option<String>,
+    pub features: Option<Vec<String>>,
```
brayniac, line 29: Should we consider a `HashSet<String>` instead? Otherwise LGTM

brayniac (approved):

PR #93, opened 2023-10-25
feat: track latencies for syscall famlies

brayniac:
Add latency distributions for each syscall family so we can better
understand the latencies for common syscalls for a given workload.


thinkingfish (approved):
nice!

PR #94, opened 2023-10-25
feat: prometheus histograms

brayniac:
Adds exposition of the full distribution of each histogram to the prometheus endpoint (`/metrics`).

This enables aggregation and display of the histogram bucket data.

This feature defaults to disabled due to the large number of additional timeseries.


src/exposition/http.rs
```diff
@@ -107,6 +121,40 @@ mod handlers {
                         }
                     }
                 }
+                if config.general().prometheus_histograms() {
+                    if let Some(snapshot) = snapshots.previous.get(metric.name()) {
+                        // downsample to 6.25% error
+                        let snapshot = snapshot.downsample(3).unwrap();
```
thinkingfish, line 127: Do you want to make the downsampling part of the config?

brayniac: Added to the config. Moved prometheus config to its own section.

src/exposition/http.rs
```diff
@@ -107,6 +122,58 @@ mod handlers {
                         }
                     }
                 }
+                if config.prometheus().histograms() {
+                    if let Some(snapshot) = snapshots.previous.get(metric.name()) {
+                        // downsample the snapshot if necessary
+                        let downsampled: Option<Snapshot> =
+                            if config.prometheus().histogram_downsampling_factor() > 0 {
```
thinkingfish, line 129: this comparison should also consider the original grouping factor not just if down-sampling factor is set

src/exposition/http.rs
```diff
@@ -107,6 +122,58 @@ mod handlers {
...
+
+                        // we need to export a total count (free-running)
+                        let mut count = 0;
+                        // we also need to export a total sum of all observations
+                        // which is also free-running
+                        let mut sum = 0;
+
+                        let mut entry = format!("# TYPE {name}_distribution histogram\n");
+                        for bucket in snapshot {
+                            // add this bucket's sum of observations
```
thinkingfish, line 156: Is it common to export bucket values as a count instead of a gauge (after delta)? Percentiles are always exported as gauges, while raw buckets can be exported as either, gauges seem more convenient at query time.

src/exposition/http.rs
brayniac: Why is that? The downsampling factor is how much we reduce the grouping power. We check that the downsample factor is valid (in relation to the grouping power) when we load the config.

src/exposition/http.rs
brayniac: It appears that is how the bucket values are expected to behave: https://prometheus.io/docs/concepts/metric_types/#histogram

It does allow easier downsampling in the time domain if the TSDB also uses free-running counters.

src/exposition/http.rs
thinkingfish: I see. `downsample` already checks that. But if the config is set to be higher than the original histogram, it would return an error during exposition. Seems something that should be verified during initialization if we don't want to silently swallow that misconfiguration?

src/exposition/http.rs
thinkingfish: Gotcha. This wasn't my intuition but I see the point.

thinkingfish (approved):

Cargo.toml
```diff
@@ -19,11 +19,11 @@ humantime = "2.1.0"
 lazy_static = "1.4.0"
 libc = "0.2.147"
 linkme = "0.3.15"
-metriken = "0.3.0"
+metriken = "0.3.2"
```
mihirn, line 22: `0.3.3` for metriken

brayniac: fixed.

mihirn (approved):

PR #128, opened 2023-12-01
add optional compression for http exposition

brayniac:
The metrics endpoint exposes a large volume of data, particularly
when histogram exposition is enabled. This can cause a substantial
amount of bandwidth to be used, especially when sampling at high
frequency.

Fortunately, the data compresses well. This change allows opt-in
compression for the endpoints using gzip.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,6 +46,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


mihirn:
This looks fine to me, but few questions:
- Do we need to change the client/scrapers in any way or is warp gzip compression totally transparent?
- Is the CPU overhead better or worse for compression vs transferring more data?
- Independent of compression, should we look at other representations of histograms at the rezolus-level or do you think it is better to keep it simple here for lower overhead and allow the consumers to reshape the data?

mihirn (approved):

brayniac:
> This looks fine to me, but few questions:
> 
> * Do we need to change the client/scrapers in any way or is warp gzip compression totally transparent?

The gzip compression is totally transparent for prometheus collector. If you need to grab the data with curl, you need to use `--compressed` flag. I'd expect most users will not enable this, but it reduced TX bandwidth from a very wide spread of ~25-50Mbps to ~2.5Mbps across my test cluster. Packet TX rate dropped from ~2.2kpps to ~200pps. 

> * Is the CPU overhead better or worse for compression vs transferring more data?

CPU overhead is higher with compression. I didn't test deflate to see if it's any better or worse. Went from 0.1-0.2 cores userspace to 0.2-0.3 cores userspace. System CPU time did drop slightly, but not enough to offset this. 

> * Independent of compression, should we look at other representations of histograms at the rezolus-level or do you think it is better to keep it simple here for lower overhead and allow the consumers to reshape the data?

The issue is that Prometheus specifies a specific way of representing the histograms. It's very inefficient compared to other representations.

mihirn:
Ack, thanks.

PR #147, opened 2024-01-24
feat: allow rezolus to run on macos

brayniac:
Allows rezolus to run on macos by cfg gating all linux-only samplers.


src/samplers/cpu/mod.rs
```diff
@@ -5,6 +5,8 @@ sampler!(Cpu, "cpu", CPU_SAMPLERS);
...
 
 #[cfg(target_os = "linux")]
-mod perf;
-
-mod proc_stat;
+#[path = "."]
+mod linux {
+    mod perf;
+    mod proc_stat;
+}
```
swlynch99, line 12: This is gonna be confusing for people coming into the code. Why not juts move the perf and proc_state files into a `linux` subfolder?

src/samplers/memory/mod.rs
```diff
@@ -5,6 +5,8 @@ sampler!(Memory, "memory", MEMORY_SAMPLERS);
...
 
 #[cfg(target_os = "linux")]
-mod proc_meminfo;
-#[cfg(target_os = "linux")]
-mod proc_vmstat;
+#[path = "."]
+mod linux {
+    mod proc_meminfo;
+    mod proc_vmstat;
+}
```
swlynch99, line 12: Same here.

src/samplers/tcp/mod.rs
```diff
@@ -4,17 +4,18 @@ sampler!(Tcp, "tcp", TCP_SAMPLERS);
 
 mod stats;
 
-mod connection_state;
-mod snmp;
+#[cfg(target_os = "linux")]
+#[path = "."]
+mod linux {
```
swlynch99, line 9: Ditto.

swlynch99 (approved):
LGTM other than I find the `#[path = "."]` parts confusing and think you should just rearrange the files to match the module structure.

PR #151, opened 2024-02-08
msgpack based exposition

brayniac:
Adds two new HTTP endpoints `/metrics/binary/metadata` and `/metrics/binary/readings` that use msgpack to represent the metric metadata and all the current values (respectively).

Combined, this allows for more efficient scraping of the metrics, particularly when full histogram distributions are desired. These endpoints should be preferred when gathering telemetry at high resolution as they avoid duplicating metric metadata for each scrape.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ humantime = "2.1.0"
 lazy_static = "1.4.0"
 libc = "0.2.147"
 linkme = "0.3.15"
-metriken = { version = "0.5.1" }
+metriken =  "0.5.1"
 memmap2 = "0.5.10"
 once_cell = "1.18.0"
 ouroboros = "0.17.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,11 +19,12 @@ humantime = "2.1.0"
 lazy_static = "1.4.0"
 libc = "0.2.147"
 linkme = "0.3.15"
-metriken = "0.3.3"
+metriken = { version = "0.5.1" }
 memmap2 = "0.5.10"
 once_cell = "1.18.0"
 ouroboros = "0.17.2"
-ringlog = "0.3.2"
+ringlog = { version = "0.6.0" }
+rmp-serde = "1.1.2"
 serde = { version = "1.0.185", features = ["derive"] }
 serde_json = "1.0.105"
 sysconf = "0.3.4"
@@ -32,6 +33,9 @@ tokio = { version = "1.32.0", features = ["full"] }
 toml = "0.7.6"
 walkdir = "2.3.3"
 warp = { version = "0.3.6", features = ["compression"] }
+serde_repr = "0.1.18"
+histogram = { version = "0.9.0", features = ["serde"] }
+chrono = { version = "0.4.33", features = ["serde"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 libbpf-rs = { version = "0.21.2", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ metriken =  "0.5.1"
 memmap2 = "0.5.10"
 once_cell = "1.18.0"
 ouroboros = "0.17.2"
-ringlog = { version = "0.6.0" }
+ringlog = "0.6.0"
 rmp-serde = "1.1.2"
 serde = { version = "1.0.185", features = ["derive"] }
 serde_json = "1.0.105"
--- a/src/exposition/http.rs
+++ b/src/exposition/http.rs
@@ -348,6 +348,8 @@ mod handlers {
     }
 }
 
+/// This struct contains all the metric metadata organized by type. This allows
+/// the creation of a schema for metric storage based on this metadata alone.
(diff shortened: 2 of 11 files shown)


src/exposition/http.rs
```diff
@@ -256,3 +347,63 @@ mod handlers {
         }
     }
 }
+
+#[derive(Serialize, Deserialize)]
+pub struct Metadata {
```
brayniac, line 352: These structs can probably be moved either a metriken crate? They will be shared with rpc-perf. A lot of the exposition code should also move into a common location.

Cargo.toml
```diff
@@ -19,11 +19,12 @@ humantime = "2.1.0"
 lazy_static = "1.4.0"
 libc = "0.2.147"
 linkme = "0.3.15"
-metriken = "0.3.3"
+metriken = { version = "0.5.1" }
```
mihirn, line 22: Is this the same as `metriken = "0.5.1"`?

src/exposition/http.rs
```diff
@@ -248,6 +270,75 @@ mod handlers {
         Ok(content)
     }
 
+    pub async fn binary_metadata() -> Result<impl warp::Reply, Infallible> {
+        let mut metadata = Metadata::new();
```
mihirn, line 274: Does this ever change during execution and if not, do we want to regenerate it every time? Seems like we could create the metadata once at start of day.

src/exposition/http.rs
```diff
@@ -248,6 +270,75 @@ mod handlers {
...
+
+        for metric in &metriken::metrics() {
```
mihirn, line 276: This should be true but just verifying: does iteration order remains consistent across lifetime of the application?

src/exposition/http.rs
```diff
@@ -256,3 +347,63 @@ mod handlers {
...
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct MetricMetadata {
+    name: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub enum MetricKind {
```
mihirn, line 374: Is this used anywhere?

src/exposition/http.rs
```diff
@@ -256,3 +347,63 @@ mod handlers {
         }
     }
 }
+
+#[derive(Serialize, Deserialize)]
+pub struct Metadata {
+    counters: Vec<MetricMetadata>,
+    gauges: Vec<MetricMetadata>,
+    histograms: Vec<(MetricMetadata, histogram::Config)>,
```
mihirn, line 355: Not a big deal, but the snapshot also contains the config, so this data is being duplicated.

src/exposition/http.rs
```diff
@@ -248,6 +270,75 @@ mod handlers {
         Ok(content)
     }
 
+    pub async fn binary_metadata() -> Result<impl warp::Reply, Infallible> {
+        let mut metadata = Metadata::new();
+
+        for metric in &metriken::metrics() {
+            let any = match metric.as_any() {
```
mihirn, line 277: Not necessarily for this patch, but I think the checks for inclusion should be centralized since if we accidentally go out of sync between metadata and data, it will be bad,

src/exposition/http.rs
brayniac: I think the metadata should contain all the information to build the schema. If anything, I would drop the config from the snapshot, but that's not possible with the current interfaces to `histogram::Snapshot`.

src/exposition/http.rs
brayniac: Good catch. Removed.

src/exposition/http.rs
brayniac: Iteration order is now consistent since we guarantee we are only exposing static metrics, the ordering in that slice is determined at link time. Comments added.

src/exposition/http.rs
brayniac: I don't think there's a huge win in caching this. It should not change, but it is relatively cheap to produce.

Cargo.toml
brayniac: Yes, cleaned up,. Artifact from me adding a `path = "..."` during development

src/exposition/http.rs
brayniac: Agree. Added a TODO for this.

src/exposition/http.rs
mihirn: Seems reasonable.

Cargo.toml
```diff
@@ -19,11 +19,12 @@ humantime = "2.1.0"
 lazy_static = "1.4.0"
 libc = "0.2.147"
 linkme = "0.3.15"
-metriken = "0.3.3"
+metriken =  "0.5.1"
 memmap2 = "0.5.10"
 once_cell = "1.18.0"
 ouroboros = "0.17.2"
-ringlog = "0.3.2"
+ringlog = { version = "0.6.0" }
```
mihirn, line 26: Nit: one more leftover from testing.

mihirn (approved):

PR #153, opened 2024-02-12
refactor msgpack exposition

brayniac:
Refactor msgpack exposition and include dynamic metrics

### DIFF
--- a/src/exposition/http.rs
+++ b/src/exposition/http.rs
@@ -264,7 +264,7 @@ mod handlers {
     pub async fn msgpack() -> Result<impl warp::Reply, Infallible> {
         let mut readings = MetricsSnapshot::new();
 
-        // iterate through the static metrics and build-up the readings object
+        // iterate through the metrics and build-up the snapshot
         for metric in &metriken::metrics() {
             let any = match metric.as_any() {
                 Some(any) => any,
--- a/src/exposition/http.rs
+++ b/src/exposition/http.rs
@@ -40,8 +40,7 @@ mod filters {
         prometheus_stats(config.clone())
             .or(human_stats())
             .or(hardware_info())
-            .or(binary_metadata())
-            .or(binary_readings())
+            .or(msgpack())
     }
 
     /// GET /metrics
@@ -70,20 +69,12 @@ mod filters {
             .and_then(handlers::hwinfo)
     }
 
-    /// GET /metrics/binary/metadata
-    pub fn binary_metadata(
+    /// GET /metrics/binary
+    pub fn msgpack(
     ) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
-        warp::path!("metrics" / "binary" / "metadata")
+        warp::path!("metrics" / "binary")
             .and(warp::get())
-            .and_then(handlers::binary_metadata)
-    }
-
-    /// GET /metrics/binary/readings
-    pub fn binary_readings(
-    ) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
-        warp::path!("metrics" / "binary" / "readings")
-            .and(warp::get())
-            .and_then(handlers::binary_readings)
+            .and_then(handlers::msgpack)
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/exposition/http.rs
```diff
@@ -270,65 +261,11 @@ mod handlers {
...
-        // consistency.
+    pub async fn msgpack() -> Result<impl warp::Reply, Infallible> {
+        let mut readings = MetricsSnapshot::new();
 
         // iterate through the static metrics and build-up the readings object
-        // NOTE: dynamic metrics are omitted from the binary outputs, this is to
-        // ensure that the iteration order is consistent through the lifetime of
-        // the process.
-        for metric in metriken::metrics().static_metrics() {
+        for metric in &metriken::metrics() {
```
mihirn, line 268: Nit: I guess comment should just say metrics, not static metrics.

src/exposition/http.rs
```diff
@@ -391,15 +330,15 @@ pub struct MetricMetadata {
...
 #[derive(Serialize, Deserialize)]
-pub struct Readings {
+pub struct MetricsSnapshot {
     datetime: DateTime<Utc>,
     unix_ns: u128,
-    counters: Vec<u64>,
-    gauges: Vec<i64>,
-    histograms: Vec<Option<Snapshot>>,
+    counters: Vec<(MetricMetadata, Option<u64>)>,
+    gauges: Vec<(MetricMetadata, Option<i64>)>,
```
mihirn, line 337: Can the value of these ever be `None`? I don't see a codepath where those would get inserted -- seems to me that if we wanted them, we would need to add an else branch to the `if let Some(value)` checks above?

src/exposition/http.rs
brayniac: Fixed.

src/exposition/http.rs
```diff
@@ -341,11 +278,32 @@ mod handlers {
             }
 
             if let Some(counter) = any.downcast_ref::<Counter>() {
-                readings.counters.push(counter.value());
+                if let Some(value) = counter.value() {
+                    readings.counters.push((
+                        MetricMetadata {
```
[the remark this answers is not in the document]
brayniac: Currently, no. Metric metadata could include descriptions and/or arbitrary key-value metadata fields. I think that in the long-run we want to support those and not flatten the metric name into some pre-formatted name. I don't have a strong opinion about whether we should acommodate that now or later.

src/exposition/http.rs
brayniac: I thought for lazy gauges/counters they could be none. But I'm not seeing where that could be the case with the latest metriken version. @swlynch99 - curious about your thoughts on this.

swlynch99: Yeah lazy counters and gauges can be `None` if the counter hasn't been used yet. (See https://github.com/iopsystems/metriken/blob/main/metriken/src/lazy.rs#L122).

You could also treat this the same as a metric being added dynamically and filter out all metrics whose values are `None`.

src/exposition/http.rs
```diff
@@ -293,59 +277,27 @@ mod handlers {
...
-                });
-            } else if let Some(histogram) = any.downcast_ref::<RwLockHistogram>() {
-                metadata.histograms.push((
+            if let Some(counter) = any.downcast_ref::<Counter>() {
+                readings.counters.push((
                     MetricMetadata {
                         name: metric.formatted(metriken::Format::Simple),
                     },
-                    histogram.config(),
+                    counter.value(),
```
swlynch99, line 285: Instead of doing downcasting here you may want to use [`Metric::value`](https://docs.rs/metriken/latest/metriken/trait.Metric.html#tymethod.value) instead. You will still need to downcast for the histograms, but counters and gauges are supported.

src/exposition/http.rs
brayniac: Here value() always returns a u64 or i64 depending on the type, I don't see how to get an option anymore?

brayniac: Fixed.

src/exposition/http.rs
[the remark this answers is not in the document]
brayniac: Simplified for now.

mihirn (approved):

PR #156, opened 2024-02-17
cleanup snapshotting within rezolus

brayniac:
Simplifies the snapshotting within rezolus to rely on the metriken exposition crate to generate the histogram snapshots.


src/exposition/http.rs
```diff
@@ -126,7 +127,9 @@ mod handlers {
             } else if any.downcast_ref::<AtomicHistogram>().is_some()
                 || any.downcast_ref::<RwLockHistogram>().is_some()
             {
-                if let Some(delta) = snapshots.deltas.get(metric.name()) {
+                let simple_name = metric.formatted(metriken::Format::Simple);
```
brayniac, line 130: we should make sure this has no performance impact

brayniac: This doesn't appear to make much of a difference.

src/main.rs
```diff
@@ -27,72 +29,29 @@ impl Default for Snapshots {
...
+        let mut snapshot = metriken_exposition::Snapshotter::default().snapshot();
+        self.timestamp = snapshot.systemtime;
 
-        for metric in metriken::metrics().iter() {
-            let any = if let Some(any) = metric.as_any() {
-                any
-            } else {
-                continue;
-            };
+        let mut current: HistogramSnapshots = snapshot.histograms.drain(..).collect();
```
mihirn, line 48: I'm curious: what is the difference between `histograms.drain(..).collect()` and the `histograms.into_iter().collect()` above? Seems like they should be the same since there is no range to the drain.

src/main.rs
```diff
@@ -27,72 +29,29 @@ impl Default for Snapshots {
...
-
-            if let Some(snapshot) = snapshot {
-                if let Some(previous) = self.previous.get(&key) {
-                    self.deltas
-                        .insert(key.clone(), snapshot.wrapping_sub(previous).unwrap());
-                }
-
-                current.insert(key, snapshot);
+        for (name, previous) in &self.previous {
+            if let Some((name, current)) = current.remove_entry(name) {
```
mihirn, line 52: So I'm a little confused as to what is happening here: if we remove the entries, won't it be empty when `self.previous = current` happens on 58?

src/main.rs
```diff
@@ -27,72 +29,29 @@ impl Default for Snapshots {
...
+                let _ = current.wrapping_sub(previous);
+                deltas.insert(name, current);
```
mihirn, line 54: Shouldn't this be `self.deltas` or probably a later  `self.deltas = deltas`?

src/main.rs
brayniac: They are effectively the same and just a stylistic difference. I've corrected for consistency.

src/main.rs
brayniac: TBH I'm not sure what my reasoning was since this PR is over a month old now. Fixed.

src/main.rs
brayniac: Fixed.

mihirn (approved):

PR #168, opened 2024-03-14
refactor perf event handling

brayniac:
Refactors perf event handling so that the counters aren't all or nothing. This allows the sampler to tolerate environments where we cannot initialize one or more of the counters.


yangxi (approved):
LGTM, test on aws, works well.

PR #197, opened 2024-03-22
Add the block sampler and the network interface sampler

yangxi:
The samplers.network_interface samples network metrics from /sys/class/net/DEVICE/statistics
The sampers.block_device samples block metrics from /sys/block/DEVICE/stat

### DIFF
--- a/crates/systeminfo/src/hwinfo/mod.rs
+++ b/crates/systeminfo/src/hwinfo/mod.rs
@@ -61,4 +61,8 @@ impl HwInfo {
     pub fn get_network_interfaces(&self) -> &Vec<Interface> {
         &self.network
     }
+
+    pub fn get_storage_blocks(&self) -> &Vec<Block> {
+        &self.blocks
+    }
 }
--- a/crates/systeminfo/src/hwinfo/mod.rs
+++ b/crates/systeminfo/src/hwinfo/mod.rs
@@ -8,6 +8,7 @@ mod memory;
 mod net;
 mod node;
 mod sched_domain;
+mod storage;
 mod util;
 
 pub use self::cache::{Cache, CacheType};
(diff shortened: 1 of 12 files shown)


src/samplers/block/stats.rs
```diff
@@ -0,0 +1,89 @@
...
+    name = "block/read/bytes",
+    description = "block read bytes ",
+    formatter = block_metric_formatter
+)]
+pub static BLOCK_READ_BYTES: LazyCounter = LazyCounter::new(Counter::default);
+
+histogram!(BLOCK_READ_BYTES_HISTOGRAM, "block/read/bytes");
+
+#[metric(
+    name = "block/read/ios",
```
brayniac, line 16: I think it's better to call this "block/read/operations"

brayniac:
I'm wondering if there's another way we can achieve this. On my storage test system, this increases CPU utilization of rezolus from ~22% to ~42% and adds about 9k read syscalls/s. The total syscall load increase is about 22k/s from 8k/s -> 30k/s.

I'm hesitant to merge this with these increases.

brayniac:
The storage sampler total read/write metrics double-count when things are layered using device mapper. I'm not sure if that's the desired behavior or if we want to try to filter things so we're only tracking metrics for physical block devices.

brayniac:
Similarly the network sampler totals will double count traffic when physical interfaces are part of a bridge

crates/systeminfo/src/hwinfo/mod.rs
```diff
@@ -45,6 +48,7 @@ impl HwInfo {
             cpu_boosting: self::cpufreq::get_cpu_boosting(),
             memory: Memory::new()?,
             network: self::net::get_interfaces(),
+            blocks: self::storage::get_blocks(),
```
brayniac, line 51: I don't think we should be calling this 'blocks' we know what this means, but in my opinion it's not great in the system info json to just have a list of things called 'blocks'. Not everybody will know what that means

PR #198, opened 2024-03-22
add byte and ops counters for bpf blockio sampler

brayniac:
We want to be able to get metrics about block device IOPS and throughput. Since we already instrument the tracepoints to
get operation latency and size distributions, it's easy to add counters for aggregate block device IOPS and throughput.


brayniac:
I think this should be preferred to the sysfs based approach in #197 as it does not add any additional syscalls and has no appreciable overhead.

PR #249, opened 2024-04-30
refactor bpf histogram indexing

brayniac:
The previous BPF histogram indexing was based on a reduction of the algorithm in an older version of the histogram crate that had both an a and b component. The latest histogram crate replaced those with a single grouping power.

This change brings the indexing more in-line with the Rust version. Specific reductions and optimizations are made based on the specific values that we use throughout Rezolus.

### DIFF
--- a/src/common/bpf/histogram.h
+++ b/src/common/bpf/histogram.h
@@ -208,74 +208,18 @@ static u32 clz(u64 value) {
 }
 
 // base-2 histogram indexing function that is compatible with Rust `histogram`
-// crate with grouping power = 4. This uses 2 pages (8KB) in kernel space and
-// 7.6KB in user space and has a relative error of 6.25%
+// crate.
 //
 // See the indexing logic here:
 // https://github.com/pelikan-io/rustcommon/blob/main/histogram/src/config.rs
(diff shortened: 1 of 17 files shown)


src/common/bpf/histogram.h
```diff
@@ -203,25 +203,23 @@ static u32 clz(u64 value) {
...
-    }
-
-    u64 h = 63 - clz(value);
-    // h < r
-    if (h < 8) {
+// crate.
+//
+// See the indexing logic here:
+// https://github.com/pelikan-io/rustcommon/blob/main/histogram/src/config.rs
+static u32 value_to_index(u8 grouping_power, u64 value) {
```
thinkingfish, line 215: I feel `grouping_power` should be the second parameter and `value` should be the first, like how `log` function in most languages allow a base being specified, and that `base` is usually the second parameter.

thinkingfish (approved):
One nit-picking otherwise LGTM

src/common/bpf/histogram.h
brayniac: 👍 fixed

PR #256, opened 2024-05-08
add cpu busy time

brayniac:
Adds metrics for cpu busy time to make cpu utilization easier to calculate. This metric is not exported in the prometheus exposition because all the non-idle states can be trivially summed. However, for users of other metrics systems or consumers of the raw msgpack output, it can be more convenient to work with just busy and idle times and not worry about enumerating and summing the various other states.

### DIFF
--- a/src/samplers/cpu/linux/usage/bpf.rs
+++ b/src/samplers/cpu/linux/usage/bpf.rs
@@ -68,13 +68,15 @@ impl OnlineCores {
     }
 
     pub fn refresh(&mut self) -> Result<(), ()> {
-        self.file.rewind()
+        self.file
+            .rewind()
             .map_err(|e| error!("failed to seek to start of file: {e}"))?;
 
         let mut count = 0;
         let mut raw = String::new();
 
-        let _ = self.file
+        let _ = self
+            .file
             .read_to_string(&mut raw)
             .map_err(|e| error!("failed to read file: {e}"))?;
 
@@ -144,8 +146,8 @@ impl CpuUsage {
         skel.attach()
             .map_err(|e| error!("failed to attach bpf program: {e}"))?;
 
-        let online_cores = OnlineCores::new()
-            .map_err(|_| error!("couldn't determine number of online cores"))?;
+        let online_cores =
+            OnlineCores::new().map_err(|_| error!("couldn't determine number of online cores"))?;
 
         let cpus = match hardware_info() {
             Ok(hwinfo) => hwinfo.get_cpus(),
--- a/src/samplers/cpu/linux/usage/bpf.rs
+++ b/src/samplers/cpu/linux/usage/bpf.rs
@@ -9,7 +9,7 @@ use super::NAME;
 
 use std::io::{Read, Seek};
 
-use metriken::MetricBuilder;
+use metriken::{DynBoxedMetric, MetricBuilder};
 
 use bpf::*;
 
@@ -33,8 +33,10 @@ impl GetMap for ModSkel<'_> {
 pub struct CpuUsage {
     bpf: Bpf<ModSkel<'static>>,
     percpu_counters: Arc<PercpuCounters>,
-    sum_prev: u64,
-    percpu_sum_prev: Vec<u64>,
+    total_busy: Counter,
+    total_idle: Counter,
+    percpu_busy: Vec<DynBoxedMetric<metriken::Counter>>,
+    percpu_idle: Vec<DynBoxedMetric<metriken::Counter>>,
     counter_interval: Interval,
     distribution_interval: Interval,
     online_cores: usize,
@@ -42,7 +44,6 @@ pub struct CpuUsage {
     online_cores_interval: Interval,
 }
 
-const IDLE_CPUTIME_INDEX: usize = 5;
 impl CpuUsage {
     pub fn new(config: &Config) -> Result<Self, ()> {
         let builder = ModSkelBuilder::default();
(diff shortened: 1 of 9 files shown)


src/samplers/cpu/macos/usage/mod.rs
```diff
@@ -124,22 +124,26 @@ impl CpuUsage {
                     .offset((cpu as i32 * libc::CPU_STATE_MAX + libc::CPU_STATE_NICE) as isize)
                     as u64)
                     .wrapping_mul(self.nanos_per_tick);
+                let busy = user.wrapping_add(system.wrapping_add(idle));
```
mihirn, line 127: Shouldn't busy be user + system + nice, not user + system + idle?

src/samplers/cpu/linux/usage/mod.bpf.c
```diff
@@ -7,23 +7,24 @@
 #include <bpf/bpf_core_read.h>
 #include <bpf/bpf_tracing.h>
 
-#define COUNTER_GROUP_WIDTH 16
+#define COUNTER_GROUP_WIDTH 8
 #define MAX_CPUS 1024
 
 #define IDLE_STAT_INDEX 5
 #define IOWAIT_STAT_INDEX 6
+#define TOTAL_STATES 10
```
mihirn, line 15: Do we use this anywhere?

src/samplers/cpu/macos/usage/mod.rs
brayniac: Good catch. Fixed.

src/samplers/cpu/linux/usage/mod.bpf.c
brayniac: Removed.

src/samplers/cpu/linux/usage/bpf.rs
```diff
@@ -142,34 +163,34 @@ impl CpuUsage {
...
-
-        // get the number of nanoseconds in busy time, since idle hasn't been
-        // incremented, the busy time is the difference between our prev and
-        // current sums
-        let busy_delta = sum_now.wrapping_sub(self.sum_prev);
+        // update busy time metric
+        let busy: u64 = busy();
+        let busy_prev = CPU_USAGE_BUSY.value();
+        let busy_delta = busy.wrapping_sub(busy_prev);
+        self.total_busy.set(elapsed.as_secs_f64(), busy);
```
mihirn, line 170: I get why you're reading from `CPU_USAGE_BUSY` and then writing to `self.total_busy` so that you don't have to handle the histogram separately, but it is a bit confusing to read. Is it worth adding a `get` or `value` to the `Counter` so that we could just call `self.total_busy.value` instead?

src/samplers/cpu/linux/usage/bpf.rs
```diff
@@ -142,34 +163,34 @@ impl CpuUsage {
...
+            .percpu_busy
+            .iter_mut()
+            .zip(self.percpu_idle.iter_mut())
+            .enumerate()
+        {
+            let busy: u64 = self.percpu_counters.sum(cpu).unwrap_or(0);
+            let busy_prev = busy_counter.set(busy);
+            let busy_delta = busy.wrapping_sub(busy_prev);
+
+            let idle_delta = (elapsed.as_nanos() as u64).saturating_sub(busy_delta);
```
mihirn, line 190: Shouldn't this check if the core is online? If we don't have that information, it seems like delta and idle will be perfect duals of each other and we can just track either one.

src/samplers/cpu/linux/usage/proc_stat.rs
```diff
@@ -12,8 +12,10 @@ pub struct ProcStat {
     interval: Interval,
     nanos_per_tick: u64,
     file: File,
-    counters_total: Vec<Counter>,
-    counters_percpu: Vec<Vec<DynBoxedMetric<metriken::Counter>>>,
+    total_counters: Vec<Counter>,
+    total_busy: Counter,
+    percpu_counters: Vec<Vec<DynBoxedMetric<metriken::Counter>>>,
```
mihirn, line 17: Afaict, this definition is the same as the `PercpuCounters` for the bpf code. Shall we hoist that structure outside to `src/common/mod.rs` to sit next to `Counter`, so both code paths can use it?

src/samplers/cpu/linux/usage/proc_stat.rs
```diff
@@ -118,18 +134,33 @@ impl ProcStat {
             let header = parts.first().unwrap();
 
             if *header == "cpu" {
-                for (field, counter) in self.counters_total.iter_mut().enumerate() {
+                let mut busy: u64 = 0;
+
+                for (field, counter) in self.total_counters.iter_mut().enumerate() {
                     if let Some(Ok(v)) = parts.get(field + 1).map(|v| v.parse::<u64>()) {
-                        counter.set(elapsed, v.wrapping_mul(self.nanos_per_tick))
+                        if field < 2 || field > 4 {
```
mihirn, line 141: I think we only want to skip `idle` and `io_wait` so should be `field < 3`? Also, should we drop a comment in as to what the magic numbers correspond to?

brayniac: Changed to use comparison to two consts to make this clearer.

src/samplers/cpu/linux/usage/bpf.rs
brayniac: Yes. This is fixed now.

src/samplers/cpu/linux/usage/bpf.rs
brayniac: This is fixed too

mihirn (approved):

PR #277, opened 2024-06-14
add json stats endpoint

brayniac:
Adds a /vars.json endpoint that provides JSON formatted metrics exposition.

### DIFF
--- a/src/exposition/http.rs
+++ b/src/exposition/http.rs
@@ -212,101 +212,16 @@ mod handlers {
     }
 
     pub async fn human_stats() -> Result<impl warp::Reply, Infallible> {
-        let mut data = Vec::new();
-
-        let snapshots = SNAPSHOTS.read().await;
-
-        for metric in &metriken::metrics() {
-            let any = match metric.as_any() {
-                Some(any) => any,
-                None => {
-                    continue;
-                }
-            };
-
-            if metric.name().starts_with("log_") {
-                continue;
-            }
-
-            let simple_name = metric.formatted(metriken::Format::Simple);
-
-            if let Some(counter) = any.downcast_ref::<Counter>() {
-                data.push(format!("{}: {}", simple_name, counter.value()));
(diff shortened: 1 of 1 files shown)


src/exposition/http.rs
```diff
@@ -54,6 +55,14 @@ mod filters {
             .and_then(handlers::human_stats)
     }
 
+    /// GET /vars.json
+    pub fn json_stats(
+    ) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
```
mihirn, line 60: Not important, but just for consistency with rpc-perf should we also support `metrics.json` for json_stats?

src/exposition/http.rs
```diff
@@ -281,3 +259,55 @@ mod handlers {
...
+// gathers up the metrics into a simple format that can be presented as human
+// readable metrics or transformed into JSON
+async fn simple_stats(quoted: bool) -> Vec<String> {
+    let mut data = Vec::new();
+
+    let snapshots = crate::SNAPSHOTS.read().await;
+
+    let q = if quoted { "\"" } else { "" };
+
+    for metric in &metriken::metrics() {
```
mihirn, line 272: Is there any reason not to use the `match metric.value() { Some(Value::Counter(value)) .. ` pattern here?

I know it probably doesn't make a huge difference since we'd chatted about how the metrics are in-crate, but seems like we may as well make over the switch-over if we are making the change.

mihirn (approved):

src/exposition/http.rs
brayniac: Good idea. PR is updated to have the same paths and copies the doc comments about formats and purposes from rpc-perf. We'll want to go back and add doc comments about the other exposition endpoints too.

src/exposition/http.rs
brayniac: Leaving this open for a follow-up PR.

PR #285, opened 2024-06-27
userspace counter improvements

brayniac:
Changes the internal behavior of the userspace counter wrapper that is used to increment both a counter and histogram. This brings the behavior more in-line with how the underlying Metriken counters work and removes the potential footguns around previous value tracking.

### DIFF
--- a/src/common/counter.rs
+++ b/src/common/counter.rs
@@ -1,0 +1,49 @@
+use metriken::AtomicHistogram;
+use metriken::LazyCounter;
+
+/// A `Counter` is a wrapper type that enables us to automatically calculate
+/// percentiles for secondly rates between subsequent counter observations.
+///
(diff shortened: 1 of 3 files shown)


src/common/counter.rs
```diff
@@ -0,0 +1,49 @@
...
+            }
+        }
+
+        self.counter.set(value)
+    }
+
+    /// Updates the counter by incrementing it by some value. If this counter
+    /// has a histogram, it normalizes the increment to a secondly rate and
+    /// increments the histogram too.
+    #[allow(dead_code)]
```
mihirn, line 41: Leftover?

brayniac: No, I think this warns on macos still because we only use this method in the linux bpf sampler for cpu usage.

mihirn (approved):

PR #356, opened 2024-10-02
rezolus-recorder

brayniac:
Adds a binary for recording Rezolus metrics to a parquet file.


src/recorder.rs
```diff
@@ -0,0 +1,283 @@
...
+        if client.is_none() {
+            debug!("connecting to Rezolus...");
+
+            match TcpStream::connect(addr).await {
+                Ok(s) => {
+                    if s.set_nodelay(true).is_err() {
+                        continue;
+                    }
+
+                    debug!("performing http2 handshake...");
```
mihirn, line 199: Curious if there is actually a benefit for HTTP/2 here given that we aren't pipe-lining or worried about head of line blocking. Better compression?

src/recorder.rs
```diff
@@ -0,0 +1,283 @@
...
+                .uri(&format!("http://{addr}/metrics/binary"))
+                .body(())
+                .unwrap();
+
+            interval.tick().await;
+
+            let start = Instant::now();
+
+            if let Ok((response, _)) = sender.send_request(request, true) {
+                if let Ok(response) = response.await {
```
mihirn, line 236: Can we use the wait for the full body in the response here rather than manually handling it in chunks? (The reverse of the current rpc-perf change which switches the other way)

mihirn:
Not a blocker, but I think it would be nice to be able to launch it with a pre-set runtime, rather than requiring Ctrl-C to quit, so it is easier to script.

mihirn (approved):

src/recorder.rs
brayniac: Probably no real benefit, but we know Rezolus supports H2, so I don't see a downside to using it.

src/recorder.rs
brayniac: The h2 client doesn't offer this afaik. It's more oriented towards being able to handle streaming responses.

brayniac:
> Not a blocker, but I think it would be nice to be able to launch it with a pre-set runtime, rather than requiring Ctrl-C to quit, so it is easier to script.

 I'll follow-up with another PR to add a timed collection

PR #410, opened 2025-01-11
fix: histogram config missing

brayniac:
Since the snapshotting moved into Rezolus, we missed adding the histogram config (power and group) to the metadata.

### DIFF
--- a/src/exposition/http/snapshot.rs
+++ b/src/exposition/http/snapshot.rs
@@ -128,7 +128,6 @@ impl Snapshot {
                                 let mut metadata = metadata.clone();
 
                                 metadata.insert("id".to_string(), counter_id.to_string());
-                                metadata.insert("group_id".to_string(), metric_id.to_string());
 
                                 if let Some(m) = counters.load_metadata(counter_id) {
                                     for (k, v) in m {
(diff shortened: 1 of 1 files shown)


src/exposition/http/snapshot.rs
```diff
@@ -116,10 +125,10 @@ impl Snapshot {
                                 if *counter == 0 {
                                     continue;
                                 }
-
                                 let mut metadata = metadata.clone();
 
                                 metadata.insert("id".to_string(), counter_id.to_string());
+                                metadata.insert("group_id".to_string(), metric_id.to_string());
```
mihirn, line 131: I actually don't think I need this, the `id->counter_id` in the line above is what I was looking for and added in `32763b57`.

src/exposition/http/snapshot.rs
```diff
@@ -144,6 +153,7 @@ impl Snapshot {
                                 let mut metadata = metadata.clone();
 
                                 metadata.insert("id".to_string(), gauge_id.to_string());
+                                metadata.insert("group_id".to_string(), metric_id.to_string());
```
mihirn, line 156: Same as previous.

src/exposition/http/snapshot.rs
brayniac: Removed.

src/exposition/http/snapshot.rs
brayniac: Removed

PR #416, opened 2025-01-16
Rewrite packaging to use a single script in a docker container

swlynch99:
This should avoid the problems with running github actions within an older docker container. It should also make building packages by hand somewhat easier.

### DIFF
--- a/debian/package.sh
+++ b/debian/package.sh
@@ -70,6 +70,26 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+pushgroup() {
+    if [ -n "$CI" ]; then
+        echo "::group::$*"
+    fi
+}
+
+nextgroup() {
+    if [ -n "$CI" ]; then
+        echo "::endgroup::"
+        echo "::group::$*"
+    fi
+}
+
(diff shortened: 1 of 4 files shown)


debian/package.sh
```diff
@@ -0,0 +1,149 @@
...
+    --chown         Change the ownership of the resulting package files.
+
+USE WITH DOCKER:
+    This script is intended to be run within a debian-based docker container.
+    As an example, consider building for ubuntu focal:
+
+    docker run -it --rm \\
+        -v \$(pwd):/mnt/rezolus \\
+        -v \$(pwd)/target/debian:/mnt/output \\
+        ubuntu:focal /mnt/rezolus/debian/package.sh --release 0 --chown \$(id -u) --verbose
```
brayniac, line 37: I'd rather that we use a supported distro/release here. Consider using `noble` as the example

brayniac (approved):
Approved, but would like to change the example in the packaging script to something with a supported kernel version.

swlynch99:
Fixed. Will merge once all CI jobs pass.

PR #437, opened 2025-02-05
use forward slash as cgroup path joiner

brayniac:
Change the cgroup name label to use a forward slash as the path joiner to make hierarchy more obvious.


swlynch99 (approved):
:ship:

PR #448, opened 2025-02-15
[RFC] Add a rezolus metrics subscription endpoint

swlynch99:
We've had a few concerns with jitter and overhead with rezolus. I originally started writing this as an experiment to see how it would look to add a subscription endpoint.

Here were my goals/constraints when designing this:
- Metadata for metrics is only sent once per metric.
- There is no unbounded memory usage on the rezolus end, there will be at most 3 snapshots being stored concurrently (1 being sent, 1 buffered, and 1 being recorded). Extra snapshots will be dropped if they are not sent in time.
- Snapshots only contain the metric id and value and nothing else.
- The protocol allows for recovering from parsing errors. Individual frames are prefixed with their length, so if a frame fails to parse then it can be skipped without having to drop the stream.
- Snapshot frequency is driven by rezolus itself, although it is configured by a query parameter.

Here's how it works. There are (currently) 4 different frame types:
- An initial metadata frame with connection-level metadata.
- A metric info frame with metadata info about metrics. A metadata entry is sent once per metric (before it shows up in a snapshot) but an info frame can contain info on multiple metrics.
- A snapshot frame which contains metric ids and values.
- A frame that indicates some number of snapshots have been lost.

I have ultimately settled on using the metric names as the metric id as I found that to be more convenient. The code could be optimized a bit to reduce allocations by using a small-string library, but I have ignored that for this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -363,6 +363,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-stream"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b5a71a6f37880a80d1d7f19efd781e4b5de42c88f0722cc13bcb6cc2cfe8476"
+dependencies = [
+ "async-stream-impl",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-stream-impl"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


brayniac:
I don't see where the samplers are being refreshed as part of this. They would need to be refreshed before each snapshot frame is created. Am I missing something?

I think to make evaluation easier we need an implementation of the rezolus recorder that subscribes to the streaming endpoint. 

So far, analysis seems to be that the jitter of the Rezolus recorder is pretty good. I think we also need to add an additional timestamp in the snapshots so we have both the start of the refresh and the completion so we can do better analysis.

swlynch99:
I don't think I'm going to have time to drive this so I'm going to close it for now.

PR #450, opened 2025-02-20
use metric definitions from metriken-exposition

mihirn:


src/exposition/http/mod.rs
```diff
@@ -63,7 +61,7 @@ fn app(state: Arc<AppState>) -> Router {
 async fn msgpack(State(state): State<Arc<AppState>>) -> Vec<u8> {
     state.refresh().await;
 
-    let snapshot = Snapshot::new();
+    let snapshot = snapshot::new_snapshot();
```
brayniac, line 64: consider renaming the function so this is `snapshot::create()` or similar?

src/exposition/http/snapshot.rs
```diff
@@ -2,176 +2,126 @@ use crate::exposition::http::CounterGroup;
...
-    pub value: histogram::Histogram,
-    pub metadata: HashMap<String, String>,
-}
-
-/// Contains a snapshot of metric readings.
-#[derive(Clone, Serialize, Deserialize)]
-#[non_exhaustive]
-pub struct Snapshot {
-    pub systemtime: SystemTime,
+pub fn new_snapshot() -> Snapshot {
```
brayniac, line 9: same as above

brayniac (approved):
Non-blocking suggestion wrt the function name. LGTM and can ship either with or without that change.

PR #461, opened 2025-03-06
add summarize subcommand

brayniac:
Adds a new subcommand 'summarize' which serves as a proxy which scrapes the msgpack endpoint periodically and produces summary metrics (percentiles) from the histograms.

This PR also adds the `rezolus-summarize` service definition which by default will connect to the default Rezolus port (4242) on localhost and listen on port 4243 on all addresses. This service is disabled by default. This is configured for secondly scraping and the interval should be set to match the scrape interval of metrics collection.

In environments where collection of percentiles instead of full histograms is desired, the `rezolus-summarize` service should be enabled, and the metrics collection should be pointed at port 4243 instead of the Rezolus agent's default port 4242.

### DIFF
--- a/src/summarize.rs
+++ b/src/summarize.rs
@@ -159,13 +159,8 @@ pub fn run(config: Config) {
 
     // timed loop to calculate summary metrics
     rt.block_on(async move {
-        // get an aligned start time
-        let start = tokio::time::Instant::now()
-            - Duration::from_nanos(Utc::now().nanosecond() as u64)
-            + config.interval.into();
-
         // sampling interval
-        let mut interval = tokio::time::interval_at(start, config.interval.into());
+        let mut interval = crate::common::aligned_interval(config.interval.into());
 
         // previous snapshot
         let mut previous = None;
--- a/src/summarize.rs
+++ b/src/summarize.rs
@@ -1,0 +1,436 @@
+use tower_http::decompression::RequestDecompressionLayer;
+use tower_http::compression::CompressionLayer;
+use tower::ServiceBuilder;
+use axum::routing::get;
+use tokio::net::TcpListener;
+use axum::Router;
+use parking_lot::Mutex;
+use std::collections::HashMap;
+use metriken_exposition::*;
+use std::time::SystemTime;
+use metriken_exposition::SnapshotV2;
+use std::net::SocketAddr;
+use metriken_exposition::Snapshot;
+use super::*;
+use clap::ArgMatches;
+
+static SUMMARIZED: Mutex<Option<SnapshotV2>> = Mutex::new(None);
+
+pub struct Config {
+    interval: humantime::Duration,
+    verbose: u8,
+    url: Url,
+    listen: SocketAddr,
+}
+impl TryFrom<ArgMatches> for Config {
+    type Error = String;
+
+    fn try_from(
+        args: ArgMatches,
+    ) -> Result<Self, <Self as std::convert::TryFrom<clap::ArgMatches>>::Error> {
+        Ok(Config {
+            url: args.get_one::<Url>("URL").unwrap().clone(),
+            listen: *args.get_one::<SocketAddr>("LISTEN").unwrap(),
+            verbose: *args.get_one::<u8>("VERBOSE").unwrap_or(&0),
+            interval: *args
+                .get_one::<humantime::Duration>("INTERVAL")
+                .unwrap_or(&humantime::Duration::from_str("1s").unwrap()),
+        })
+    }
+}
+
+pub fn command() -> Command {
+    Command::new("summarize")
+        .about("Produce and expose summary metrics for a running Rezolus agent")
+        .arg(
+            clap::Arg::new("URL")
+                .help("Rezolus HTTP endpoint")
+                .action(clap::ArgAction::Set)
+                .value_parser(value_parser!(Url))
+                .required(true)
+                .index(1),
+        )
+        .arg(
+            clap::Arg::new("LISTEN")
+                .help("Path to the output file")
+                .action(clap::ArgAction::Set)
+                .value_parser(value_parser!(SocketAddr))
+                .required(true)
+                .index(2),
+        )
+        .arg(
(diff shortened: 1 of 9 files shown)


src/summarize.rs
```diff
@@ -0,0 +1,430 @@
...
+            clap::Arg::new("URL")
+                .help("Rezolus HTTP endpoint")
+                .action(clap::ArgAction::Set)
+                .value_parser(value_parser!(Url))
+                .required(true)
+                .index(1),
+        )
+        .arg(
+            clap::Arg::new("LISTEN")
+                .help("Path to the output file")
```
mihirn, line 55: Wrong parameter description

src/summarize.rs
```diff
@@ -0,0 +1,430 @@
...
+    let config = Arc::new(config);
+
+    // launch the HTTP listener
+    let c = config.clone();
+    rt.spawn(async move { serve(c).await });
+
+    // timed loop to calculate summary metrics
+    rt.block_on(async move {
+        // get an aligned start time
+        let start = tokio::time::Instant::now()
```
mihirn, line 163: I'm still not sure this is the right ordering between `Instant::now()` and `Utc::now()`

src/summarize.rs
```diff
@@ -0,0 +1,430 @@
...
+                }
+            } else {
+                error!("failed to get metrics. terminating early");
+                break;
+            }
+        }
+    })
+}
+
+pub fn systemtime(snapshot: &Snapshot) -> SystemTime {
```
mihirn, line 213: Since these are duplicated from metriken-exposition, should we just make those publicly accessible and release a new version instead of duplicating?

src/summarize.rs
```diff
@@ -0,0 +1,430 @@
...
+    let mut summarized = summarized.unwrap();
+
+    let timestamp = summarized
+        .systemtime
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap()
+        .as_millis();
+
+    for mut metric in summarized.counters.drain(..) {
+        let name = metric.name.clone();
```
mihirn, line 375: 375-391 feels like it could be a function which takes the type (counter or gauge) and the value (as a string, since they are u64 vs i64) and produces a formatted string to push into data, so that 399-417 can also use it.

brayniac:
> The code itself looks fine to me, but I do wonder if this should be a part of the recorder or a configurable feature of Rezolus itself, instead of a separate service.

Why would you want to try to put it in the recorder or the agent?

The agent is currently totally unaware of the external sampling intervals and I'm not sure we want to have it doing work if metrics are not being collected. 

The recorder and flight-recorder are both targeted at producing file artifacts. I'm not sure that a metrics proxy fits neatly in those either.

I'm curious what the reasoning is to converge towards a single running process instead of having separation of duties.

src/summarize.rs
brayniac: Yes, unfortunately the current metriken-exposition didn't expose these in a snapshot version independent way. This should be moved into that crate and a new version published.

src/summarize.rs
brayniac: Fixed

src/summarize.rs
brayniac: See latest commit. I've moved all the alignment into a shared function with some error bounds. I'm not sure if the ordering really matters, what we care about is that there isn't a significant amount of elapsed time between those two times.

src/summarize.rs
brayniac: We can't operate on those uniformly without their being a unifying trait. I've added a local trait to hide those edges. `metriken-exposition` should try to make this easier maybe?

brayniac:
> Valid point about the agent not being aware of external sampling interval -- that seems like a pretty good reason for this to not be part of that. For the rest, I've broadly got two concerns:
> 
> a) I'm not too sure what the behaviour will be if we end up having both recorder and summarize running? As I understand it, snapshots from the binary-endpoint are demand driven, so we'll be doing unnecessary work, but since everything is free-running, at least the data should be valid?
> 
> It does seem to me like having something where there is a single service scraping the endpoint and then doing all the necessary expositions with it (file, prometheus, both) ends up making it easier to reason about who is scraping that endpoint and when.
> 
> b) There's a fair amount of duplication around formatting and I do worry a little bit that something will get out-of-sync and we'll get different looking results whether we scrape prometheus directly from the agent or through the summarize service. I guess one way of working around that would be to remove the prometheus endpoint from the agent altogether and always have all prometheus exposition through summarize (in a renamed form, since it can either be summaries or full histograms).
> 
> I don't understand the snapshot mechanism of rezolus well-enough to know the downsides of either of those, but that's broadly what my thinking was.

I think it's very likely that we will have multiple consumers - although not always from the msgpack endpoint. If we're encouraging the use of flight-recorder in production with high temporal resolution compared to standard obs collection intervals, then we have two consumers - one is definitely msgpack, and the other is either msgpack or prom depending on whether it's through the summarize proxy or not. 

Since the values are free-running, I don't think there's much concern in terms of correctness or who is reading when. We refresh the internal view and serialize it into either format. Sampler refresh is largely copying values out of mmap()'d regions since much of the telemetry is eBPF based now. We could implement some sort of TTL based freshness to skip unnecessary refreshes if we're concerned about for instance reading memory info from /proc/meminfo or refreshing our perf counter views.

I'm not sure that we can or should be concerned about who is reading and when? Is there something specific I'm overlooking?

I think the formatting logic needs to get moved out into a common location. That's a valid concern. We could move that all into a prom exporter service that is a msgpack consumer and exposes either histograms or summaries. That might help clarify the name for this service too.

Would that address the concerns you have? Happy to discuss on slack too.

mihirn:
I was mostly worried about blocking if concurrent refreshes happen, but maybe that isn't a concern since the eBPF internal snapshot happens separately and this refresh is read-only and multiple can happen in parallel? If so, most of my concerns around multiple samplers disappear.

Yeah, I think moving all the exporter stuff into a separate service is probably cleaner and easier long-term.

brayniac:
Closing in favor of #470

PR #462, opened 2025-03-07
track softirq cpu usage with its own tracepoints

brayniac:
The cpu_usage accounting for softirq time is a bit inaccurate due to the tick based nature of the kernel function we probe.

To address the undercounting of softirq time, we add tracepoints for the softirq entry and exit and calculate time spent in the softirq handlers.

This makes it very low-cost to add additional metrics for tracking both softirq counts and softirq time by category. This will provide additional insight into the cause of high softirq time.


yangxi (approved):
LGTM

PR #470, opened 2025-03-24
add rezolus exporter subcommand

brayniac:
Adds an exporter which acts as a client of the Rezolus agent and provides a prometheus compatible endpoint.

To be able to create summary metrics from histograms, periodic snapshots must be taken to produce delta histograms from the otherwise free-running histograms. To keep the Rezolus agent as a passive component, the exposition is moved into a separate subcommand so that periodic snapshotting can be introduced.

This change introduces the new exporter subcommand, which by default will listen on port :4241 and connect to a Rezolus agent on localhost port :4242.


src/agent/samplers/syscall/linux/mod.rs
```diff
@@ -26,7 +26,7 @@ pub fn syscall_lut() -> Vec<u64> {
                     // sleep
                     "clock_nanosleep" | "nanosleep" => 6,
                     // socket creation and management
-                    "accept" | "bind" | "connect" | "getpeername" | "getsockname"
+                    "accept" | "accept4" | "bind" | "connect" | "getpeername" | "getsockname"
```
mihirn, line 29: Does anything else need to happen to support this metric or is adding it to the list sufficient?

src/exporter/config/prometheus.rs
```diff
@@ -0,0 +1,61 @@
...
+    // control whether full histograms should be exported and what grouping
+    // power to export them at
+    #[serde(default = "disabled")]
+    histograms: bool,
+    #[serde(default = "histogram_grouping_power")]
+    histogram_grouping_power: u8,
+
+    // the exporter samples periodically, this controls that interval
+    #[serde(default = "interval")]
+    interval: String,
```
mihirn, line 14: I wonder if the interval should go outside prometheus, in case we ever want to add another sampling endpoint to the exporter. Arguably, that could have a separate sampling interval, so I could go either way on this.

src/exporter/config/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+fn enabled() -> bool {
+    true
+}
+
+fn histogram_grouping_power() -> u8 {
+    HISTOGRAM_GROUPING_POWER
+}
+
+fn listen() -> String {
+    "0.0.0.0:4243".into()
```
mihirn, line 32: Should this be 4241 to match the docs/default toml config?

src/agent/samplers/syscall/linux/mod.rs
brayniac: This wasn't intended to be in this PR, but would indeed be all that needs to happen for that syscall to count in that group. Let me know if you prefer I split that out into a separate PR.

src/exporter/config/mod.rs
brayniac: Yes. I was having port number debates with myself and the point you mention below was a consideration. We definitely want it to be consistent

src/exporter/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        )
+}
+
+async fn root() -> String {
+    let version = env!("CARGO_PKG_VERSION");
+    format!("Rezolus {version} Exporter\nFor information, see: https://rezolus.com\n")
+}
+
+// for convenience, this proxies the msgpack from Rezolus Agent
+async fn msgpack(State(state): State<Arc<AppState>>) -> Vec<u8> {
```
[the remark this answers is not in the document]
brayniac: I think that would work and is probably the least intrustive in terms of upgrade path. I'll switch those around.

src/common/mod.rs
```diff
@@ -47,3 +49,12 @@ pub fn cpus() -> Result<Vec<usize>, Error> {
 
     Ok(ids)
 }
+
+pub fn aligned_interval(interval: Duration) -> tokio::time::Interval {
```
mihirn, line 53: I could have sworn you had a version of this which did some sanity checking to make sure there hadn't been an abnormal pause in between the `Instant::now()` and the `Utc::now()` and that the start was still in the future. Do you remember anything like that or am I imagining it?

brayniac: You are right! I missed copying that logic over from the other branch. Let me fix that.

src/exporter/snapshot.rs
```diff
@@ -0,0 +1,132 @@
+use super::*;
+
+/// Produces a snapshot from a previous and current snapshot
+pub fn snapshot(config: &Config, mut previous: Snapshot, mut current: Snapshot) -> SnapshotV2 {
+    let mut snapshot = SnapshotV2 {
+        systemtime: current.systemtime(),
+        duration: current
```
mihirn, line 7: I thought we were using the duration field for how long the snapshot took, so this shouldn't this be the latency rather than the interval between snapshots (which we'll get when we delta systemtime between them anyway)?

brayniac: I thought we wanted it to reflect the time it takes to generate the snapshot. We wouldn't need to encode the duration between consecutive snapshots, since that can be taken from the deltas of the systemtime column. I thought we decided we wanted to know when snapshotting takes an unusually large amount of time:

See current usage here where it is the latency of refreshing the samplers: https://github.com/iopsystems/rezolus/blob/main/src/agent/exposition/http/mod.rs#L68

src/exporter/config/mod.rs
brayniac: Port numbers should hopefully be consistent now

src/exporter/config/prometheus.rs
brayniac: I was torn on this, but let me move that up into `general` section. If we allow per exposition format settings later, it still feels like we should have this default at a higher level in the config.

src/exporter/snapshot.rs
mihirn: I agree that it should be the time that snapshot creation takes. Maybe I'm reading it wrong, but wouldn't `current.systemtime.duration_since(previous.systemtime())` not give you that? It seems to me like it should either be the sampling latency from `mod:122` or reflect the duration field from current.

src/exporter/config/prometheus.rs
brayniac: And this is now moved into general.

src/exporter/snapshot.rs
brayniac: Ah, I see now. Yes, I'm not sure if the latency at mod:122 is the correct value to use as that includes network latency. If we can guarantee it's a v2 snapshot we read from Rezolus, then we can just use the duration field from there. Maybe we should take the field from the current snapshot if available, otherwise use the latency as measured from the exporter?

mihirn: I think that makes sense. My guess is that most scraping will be local so the network latency shouldn't hugely distort it, but we may as well use the real value if it is available.

brayniac: Latest diff addresses this by using the duration from snapshot or the latency as measured by the exporter.

mihirn: LGTM

PR #487, opened 2025-04-15
add l3 cache sampler

brayniac:
Adds a sampler that tracks l3 cache accesses and misses. Currently supports AMD Zen processors.


yangxi (approved):
LGTM

PR #499, opened 2025-04-25
fix debian packaging and hindsight service

brayniac:
Fixes the packaging to move all service registration into the rules file.

Fixes a few small issues using Rezolus Hindsight with the packaged config.


debian/rules
```diff
@@ -35,4 +35,8 @@ override_dh_auto_build:
 
 override_dh_auto_install:
 	@mkdir -p debian/tmp/bin
-	cp -f "$(BINDIR)/rezolus" debian/tmp/bin/
\ No newline at end of file
+	cp -f "$(BINDIR)/rezolus" debian/tmp/bin/
+
+override_dh_installsystemd:
+	dh_installsystemd --name=rezolus
+	dh_installsystemd --name=rezolus-exporter
```
swlynch99, line 42: I think you need to scope this to the specific package, and also make sure to run it for other packages
```suggestion
	dh_installsystemd -p rezolus --name=rezolus
	dh_installsystemd -p rezolus --name=rezolus-exporter
	dh_installsystemd --remaining-packages
```

swlynch99 (approved):

PR #505, opened 2025-05-05
use async http client

brayniac:
Previously, we were using the blocking Reqwest client in an async context. While done intentionally, this does result in a panic in debug builds.

Changes the recorder, exporter, and hindsight tools to use an async client in the async context.

Updates the ci smoketest to be a combination of the agent and recorder.


src/hindsight/mod.rs
```diff
@@ -89,8 +89,17 @@ pub fn run(config: Config) {
 
     let url = config.general().url();
 
-    // our http client
-    let client = match Client::builder().http1_only().build() {
+    // our blocking http client
```
mihirn, line 92: I'm guessing having a blocking client here is fine since it isn't an async function?

brayniac: Correct, we're not in an async runtime there, so we *must* use a blocking client.

src/hindsight/mod.rs
```diff
@@ -89,8 +89,17 @@ pub fn run(config: Config) {
...
+    let blocking_client = match reqwest::blocking::Client::builder().http1_only().build() {
+        Ok(c) => c,
+        Err(e) => {
+            error!("error connecting to Rezolus: {e}");
+            std::process::exit(1);
+        }
+    };
+
+    // our async http client
+    let async_client = match reqwest::Client::builder().http1_only().build() {
```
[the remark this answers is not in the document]
mihirn: Fair enough, not a strong opinion either way.

mihirn (approved):

PR #507, opened 2025-05-09
prevent undercounting of throttled time

brayniac:
Previously, we set the throttle start time for every cfs_throttle_rq() which can result in the throttling time being undercounted if it is updated again before unthrottling occurs. 

This checks if the throttle start and early returns if it is non-zero


brayniac:
After discussion, closing in favor of #509

PR #509, opened 2025-05-13
Fix the cgroup throttled time 

yangxi:
Each cgroup has per-CPU run queues, and the run queues enter the throttling independently, so we track the throttling start time of each run queue.

### DIFF
--- a/src/agent/samplers/cpu/linux/bandwidth/mod.bpf.c
+++ b/src/agent/samplers/cpu/linux/bandwidth/mod.bpf.c
@@ -11,6 +11,7 @@
 #include <bpf/bpf_core_read.h>
 #include <bpf/bpf_tracing.h>
 
+#define MAX_CPUS 1024
 #define MAX_CGROUPS 4096
 #define RINGBUF_CAPACITY 262144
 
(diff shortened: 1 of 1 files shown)


src/agent/samplers/cpu/linux/bandwidth/mod.bpf.c
```diff
@@ -200,8 +202,8 @@ int throttle_cfs_rq(struct pt_regs *ctx)
 
     // record throttle start time
     u64 now = bpf_ktime_get_ns();
-    u32 cgroup_idx = (u32)cgroup_id;
-    bpf_map_update_elem(&throttle_start, &cgroup_idx, &now, BPF_ANY);
+    u32 cgroup_runqueue_idx = (u32)cgroup_id * MAX_CPUS + cpu;
```
brayniac, line 205: My instinct is that CPU should be the outer key so that cachelines don't bounce around? 

Eg:
`cpu * MAX_CGROUPS + (u32)cgroup_id`

src/agent/samplers/cpu/linux/bandwidth/mod.bpf.c
```diff
@@ -235,8 +238,8 @@ int unthrottle_cfs_rq(struct pt_regs *ctx)
         return 0;
 
     // lookup start time
-    u32 cgroup_idx = (u32)cgroup_id;
-    u64 *start_ts = bpf_map_lookup_elem(&throttle_start, &cgroup_idx);
+    u32 cgroup_runqueue_idx = (u32)cgroup_id * MAX_CPUS + cpu;
```
brayniac, line 241: Same here if we adjust the code above

brayniac (approved):

PR #512, opened 2025-05-15
Add cgroup_cpu_bandwidth metrics

yangxi:
Add three cgroup_cpu_bandwidth metrics for tracking the cgroup periods, throttled periods, and throttled time.

### DIFF
--- a/src/agent/samplers/cpu/linux/bandwidth/stats.rs
+++ b/src/agent/samplers/cpu/linux/bandwidth/stats.rs
@@ -18,35 +18,35 @@ pub static CGROUP_CPU_BANDWIDTH_PERIOD_DURATION: GaugeGroup = GaugeGroup::new(MA
 
 #[metric(
     name = "cgroup_cpu_throttled_time",
-    description = "The total time a cgroup has been throttled by the CPU controller in nanoseconds",
+    description = "The total time all runqueues in a cgroup throttled by the CPU controller in nanoseconds",
     metadata = { unit = "nanoseconds" }
 )]
 pub static CGROUP_CPU_THROTTLED_TIME: CounterGroup = CounterGroup::new(MAX_CGROUPS);
 
 #[metric(
     name = "cgroup_cpu_throttled",
-    description = "The number of times a cgroup has been throttled by the CPU controller",
+    description = "The number of times all runqueues in a cgroup throttled by the CPU controller",
     metadata = { unit = "events" }
 )]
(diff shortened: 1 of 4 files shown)


src/agent/samplers/cpu/linux/bandwidth/stats.rs
```diff
@@ -10,11 +10,11 @@ use crate::agent::*;
 pub static CGROUP_CPU_BANDWIDTH_QUOTA: GaugeGroup = GaugeGroup::new(MAX_CGROUPS);
 
 #[metric(
-    name = "cgroup_cpu_bandwidth_period",
+    name = "cgroup_cpu_bandwidth_period_duration",
```
brayniac, line 13: What was the reason for renaming this? That's a pretty verbose metric name now.

src/agent/samplers/cpu/linux/bandwidth/stats.rs
```diff
@@ -29,3 +29,24 @@ pub static CGROUP_CPU_THROTTLED_TIME: CounterGroup = CounterGroup::new(MAX_CGROU
     metadata = { unit = "events" }
 )]
 pub static CGROUP_CPU_THROTTLED: CounterGroup = CounterGroup::new(MAX_CGROUPS);
+
+#[metric(
+    name = "cgroup_cpu_bandwidth",
```
brayniac, line 34: Metrics with the same name should be measuring the same thing where some key-value metadata is used to distinguish between types of that same thing. For instance: `blockio_operation` can be categorized as `read` or `write` but summing them together gives us total operations (which is also a useful way to look at that metric).

yangxi: fixed

src/agent/samplers/cpu/linux/bandwidth/stats.rs
```diff
@@ -29,3 +29,24 @@ pub static CGROUP_CPU_THROTTLED_TIME: CounterGroup = CounterGroup::new(MAX_CGROU
...
+    description = "Cgroup CPU bandwidth periods",
```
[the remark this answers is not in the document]
yangxi: fixed

src/agent/samplers/cpu/linux/bandwidth/stats.rs
yangxi: Becasue it's very similar with the name of the new metric `cgroup_cpou_bandwidth_periods and the cgroup_cpu_bandwidth_period shows the duration of a period.

brayniac (approved):

PR #517, opened 2025-05-22
Using per-task utime and stime to count the CPU usage

yangxi:
Problem

The kprobe handler is not always invoked. Missed cpuacct_account_field_kprobe causes CPU usage undercounting.

Solution

Kernel increases per-task utime/stime before invoking cpuacct_account_field. In the kprobe, we can count the user time and system time by tracking per-task utime/stime. Because the per-task utime includes both the CPUTIME_NICE and CPUTIME_USER and the stime includes CPUTIME_SYSTEM, CPUTIME_SOFTIRQ and CPUTIME_IRQ, the per-CPU and per-cgroup user time and the system time we count in the ebpf includes these time too.

Result

Under system intensive stress-ng benchmarks, readings from rezolus, /proc/stat, and perf agree with each other.

### DIFF
--- a/src/agent/samplers/cpu/linux/usage/mod.bpf.c
+++ b/src/agent/samplers/cpu/linux/usage/mod.bpf.c
@@ -10,6 +10,7 @@
 
 #define CPU_USAGE_GROUP_WIDTH 8
 #define MAX_CPUS 1024
+#define MAX_PID 4194304
 #define MAX_CGROUPS 4096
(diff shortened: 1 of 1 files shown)


brayniac (approved):

brayniac:
This looks good to me. I think we'll need to lift the irq/softirq time out of the cpu_usage metric then. I can follow-up with a separate PR to address that.

PR #548, opened 2025-06-11
feat(agent): virtualization aware threading strategy for perf events

brayniac:
Previously, the perf groups for cpu_frequency sampler were read serially and blocked the sampler. On some systems, it can take several milliseconds to collect the readings in this way (m7i.8xl). However, switching to multi-threaded collection is not a clear win on all systems. 

This change adds basic virtualization detection and implements single and multi threaded collection for both the cpu_frequency sampler and all BPF-based perf samplers.

As a result the cpu_frequency sampler latency should improve on virtualized systems, and cpu_perf sampler cpu usage should decrease on baremetal systems.

### DIFF
--- a/src/agent/bpf/builder.rs
+++ b/src/agent/bpf/builder.rs
@@ -116,11 +116,7 @@ impl PerfCounters {
                 perf_threads
                     .send(std::thread::spawn(move || {
                         if !core_affinity::set_for_current(core_affinity::CoreId { id: cpu }) {
-                            unpinned
-                                .send(counters)
-                                .expect("failed to send unpinned perf counters");
-                            pt_pending.fetch_sub(1, Ordering::Relaxed);
-                            return;
+                            warn!("failed to pin perf thread for core: {}", cpu);
                         }
 
                         pt_pending.fetch_sub(1, Ordering::Relaxed);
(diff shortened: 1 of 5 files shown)


src/common/mod.rs
```diff
@@ -80,3 +80,24 @@ pub fn utc_instant() -> (DateTime<Utc>, Instant) {
     eprintln!("could not get a UTC time and Instant pair");
     std::process::exit(1);
 }
+
+/// This function is best-effort detection of if the code is running inside of a
+/// virtual machine.
+pub fn is_virt() -> bool {
+    let sys_vendor = std::fs::read_to_string("/sys/class/dmi/id/sys_vendor")
```
mihirn, line 87: Not necessary for the moment, but just putting a comment in case this becomes critical in the future: RedHat have `virt-what` which has a potentially more robust check around virtual machine detection (at least on x86) here: http://git.annexia.org/?p=virt-what.git;a=summary

src/agent/bpf/builder.rs
```diff
@@ -75,6 +76,163 @@ impl PerfCounters {
...
+                let pt_pending = pt_pending.clone();
+
+                perf_threads
+                    .send(std::thread::spawn(move || {
+                        if !core_affinity::set_for_current(core_affinity::CoreId { id: cpu }) {
+                            unpinned
+                                .send(counters)
+                                .expect("failed to send unpinned perf counters");
+                            pt_pending.fetch_sub(1, Ordering::Relaxed);
+                            return;
```
mihirn: Given that the pinned and unpinned threads have the same loop, it isn't clear to me why we early return here and then handle the unpinned separately, as opposed to just accounting for the thread as unpinned, but falling through to the same main loop.

brayniac: That's actually not true? If unpinned we use a single thread to collect from all the CPUs we couldn't pin a thread for. For instance if pinning succeeds on cores 0-3, but fails on 4-7 - we will have 4 pinned threads and 1 unpinned thread.

mihirn: Or wait, is there a single unpinned thread handling all the core counters which couldn't be pinned, in which case, I wonder if it could use a call to `spawn_single` (which is modified to take a counter list).

mihirn: Ah, our messages crossed, but yeah, realised it was a single unpinned thread after that. Can we re-use spawn_single for that?

brayniac: We could potentially just do this as best-effort pinning. I'm wondering if it's better to have multiple threads still even if they're remote reads... I think we're unlikely to hit this case where setting affinity fails anyway.

Let's do it as best-effort pinning. That will simplify this

mihirn (approved):
LGTM, I think either re-using `spawn_single` for the unpinned or just doing best-effort are totally fine.

Issue #549, opened 2025-06-13
viewer url returns 404


chhetripradeep:
Sorry I am trying rezolus for the first time so pardon me in case it's my mistake. 

I downloaded the rezolus v5.0.0 deb and installed it. I ensured that the rezolus service is running fine via systemctl.

I recorded using the following command:
```
$ rezolus record --interval 1s --duration 1m http://0.0.0.0:4241 rezolus.parquet
2025-06-12T17:55:54.207+00:00 INFO [rezolus::recorder] recording metrics... ctrl-c to terminate early
2025-06-12T17:56:55.444+00:00 INFO [rezolus::recorder] finalizing recording... ctrl+c to terminate early
```
I tried to view:
```
$ rezolus view rezolus.parquet 0.0.0.0:10000
2025-06-12T17:58:52.425+00:00 INFO [rezolus::viewer] Loading data from parquet file...
2025-06-12T17:58:52.462+00:00 INFO [rezolus::viewer] Generating dashboards...
2025-06-12T17:58:53.495+00:00 INFO [rezolus::viewer] Use your browser to view: http://0.0.0.0:10000
```

Opening the url returns 404 but /about returns the content. I am not sure what I am missing. I would appreciate any help.

brayniac:
This should be fixed in the new 5.0.1 release. Thanks for reporting

PR #550, opened 2025-06-13
fix: viewer page not found error

brayniac:
When the Rezolus Viewer is run from a production build, we expect that all the served assets will be from copies embedded in the binary. This allows the Viewer to be run from anywhere.

Unfortunately, in 5.0.0 release, the index.html page was not served from an embedded copy. This results in a 404 unless the Viewer is run from a cloned copy of the repo.

This change fixes addresses this by serving the index from the embedded copy unless developer-mode is enabled at build time.


brayniac:
Fixes #549

PR #553, opened 2025-06-13
feat(agent): add a snapshot TTL

brayniac:
Adds a configurable snapshot TTL to the Rezolus Agent to allow serving cached snapshots if they are not older than the TTL.

This helps prevent excessive resource utilization.

### DIFF
--- a/src/agent/exposition/http/snapshot.rs
+++ b/src/agent/exposition/http/snapshot.rs
@@ -7,6 +7,7 @@ use std::collections::HashMap;
 use std::time::{Duration, SystemTime};
 
 pub struct SnapshotBuilder {
+    last: Option<Instant>,
     samplers: Arc<Box<[Box<dyn Sampler>]>>,
     snapshot: Snapshot,
     ttl: Duration,
@@ -28,6 +29,7 @@ impl SnapshotBuilder {
         };
 
         Self {
+            last: None,
             samplers,
             snapshot: Snapshot::V2(snapshot),
             ttl: config.general().ttl(),
@@ -35,6 +37,8 @@ impl SnapshotBuilder {
     }
 
     async fn refresh(&mut self) {
+        let last = Instant::now();
+
         // get start timestamp
         let timestamp = SystemTime::now();
 
@@ -53,16 +57,14 @@ impl SnapshotBuilder {
 
         // update the cached snapshot
         self.snapshot = create(timestamp, duration);
+
+        // update the timestamp
+        self.last = Some(last);
     }
 
-    pub async fn build(&mut self) -> &Snapshot {
-        let previous = match &self.snapshot {
-            Snapshot::V1(s) => s.systemtime,
-            Snapshot::V2(s) => s.systemtime,
-        };
-
-        if let Ok(duration) = previous.elapsed() {
-            if duration < self.ttl {
+    pub async fn build(&mut self, now: Instant) -> &Snapshot {
+        if let Some(last) = self.last {
+            if now.duration_since(last) < self.ttl {
                 return &self.snapshot;
             }
         }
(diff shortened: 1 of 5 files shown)


src/agent/exposition/http/snapshot.rs
```diff
@@ -6,7 +6,76 @@ use metriken_exposition::{Counter, Gauge, Histogram, Snapshot, SnapshotV2};
...
+        debug!("sampling latency: {} us", duration.as_micros());
+
+        // update the cached snapshot
+        self.snapshot = create(timestamp, duration);
+
+        // update the timestamp
+        self.last = Some(last);
+    }
+
+    pub async fn build(&mut self, now: Instant) -> &Snapshot {
```
mihirn, line 55: Do we want to the `now` to be passed in here or do we want to time the delta from when the function actually executes? It shouldn't be a huge difference in most cases, but in case of multiple callers build might be called a fair amount of time after the endpoint is hit.

src/agent/exposition/http/snapshot.rs
```diff
@@ -6,7 +6,76 @@ use metriken_exposition::{Counter, Gauge, Histogram, Snapshot, SnapshotV2};
 use std::collections::HashMap;
 use std::time::{Duration, SystemTime};
 
-pub fn create(timestamp: SystemTime, duration: Duration) -> Snapshot {
+pub struct SnapshotBuilder {
+    last: Option<Instant>,
+    samplers: Arc<Box<[Box<dyn Sampler>]>>,
+    snapshot: Snapshot,
```
mihirn: Purely an aesthetic choice, but I kind of feel like either both `last` and `snapshot` should be `Option` types or neither.

src/agent/exposition/http/snapshot.rs
brayniac: So initially I had it that way, but the following scenario came to mind:
A - arrives at the top of the second and we take the lock and start processing. Start timestamp for the snapshot is top of the second.
B - arrives < TTL after A
Sampling completes, but took more than the TTL.

We could serve the cached snapshot from A for B since their arrival times were closer than the TTL and they're both interested in the snapshot near the top of the second.

If we use the time after acquiring the lock and starting the refresh, B will see that the snapshot is expired, and return some snapshot starting at top of the second + duration of snapshot for request A. 

In this way I think using the timestamp of the request expresses the intent more? We take a new timestamp on L#40 instead of using the request time to timestamp the snapshot, so in that way I think that if it takes a while to get there and we do need to refresh the snapshot, the resulting snapshot is allowed to be served for the next TTL anyway.

src/agent/exposition/http/snapshot.rs
brayniac: I didn't love the idea of taking `Instant::now() - TTL` as the initial value for last. Unfortunately, since Instant is opaque, you can't construct an Instant where the inner fields are zero'd.

We could make the snapshot an `Option<T>` if we really want.

brayniac: Done in the latest commit.

src/agent/exposition/http/snapshot.rs
mihirn: I guess the case I had in mind was that if you hit the endpoint and then get data from quite far in the past, you're more likely to just reject it and hit the endpoint again, and that ends up increasing load. On the other hand, if you've sampled some other source before you hit the endpoint, you probably want the snapshot closed to the time you actually hit it rather than when you got the response, so it's a bit of a wash...

I suspect it is fine to keep it either way, so long as we have an understanding of the semantics and can use accordingly.

src/agent/exposition/http/snapshot.rs
mihirn: lgtm

PR #559, opened 2025-06-18
feat(agent): enable realtime scheduling on linux

brayniac:
Adds scheduler configuration to the Rezolus Agent. This allows for a scheduler policy to be set and to adjust the priority/niceness.

By default, Rezolus Agent will run as a realtime service with a priority of 15 to balance between realtime responsiveness and safety.

As a result, metrics snapshots will be produced with better timing guarantees.

### DIFF
--- a/src/agent/config/scheduler.rs
+++ b/src/agent/config/scheduler.rs
@@ -56,7 +56,6 @@ impl Scheduler {
                     std::process::exit(1);
                 }
             }
-            // Mismatched policy and parameters
             _ => {
                 eprintln!(
                     "scheduler policy {:?} is incompatible with the provided parameters",
@@ -72,8 +71,8 @@ impl Scheduler {
         self.set_scheduler();
 
         match self.parameters {
-            Parameters::Realtime { priority } => self.set_realtime_scheduler(priority),
-            Parameters::Normal { niceness } => self.set_normal_scheduler(niceness),
+            Parameters::Normal { niceness } => self.set_niceness(niceness),
+            _ => {}
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/agent/config/scheduler.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    #[cfg(target_os = "linux")]
+    fn set_realtime_scheduler(&self, priority: i32) {
+        let policy = match self.policy {
+            Policy::Fifo => SCHED_FIFO,
+            Policy::RoundRobin => SCHED_RR,
+            _ => unreachable!(),
+        } | SCHED_RESET_ON_FORK;
+
+        let param = libc::sched_param {
```
mihirn: Not a blocked by any means, but it seems to me like L86-107 (param, sched_setscheduler, check return and exit) could be abstracted into a function which takes a libc policy and a priority and is then common between the realtime and the normal code path.

The realtime code path would then just be the mapping from the Rust policy to the libc one, while the normal one would be the call, with priority 0, and the niceness call.

mihirn (approved):

src/agent/config/scheduler.rs
brayniac: Good point! I've addressed the review feedback.

src/agent/config/scheduler.rs
```diff
@@ -0,0 +1,129 @@
...
+        if result == -1 {
+            let e = std::io::Error::last_os_error();
+            eprintln!("could not set scheduler policy: {:?} error: {e}");
+            std::process::exit(1);
+        }
+    }
+
+    #[cfg(target_os = "linux")]
+    fn set_niceness(&self, niceness: i32) {
+        assert!(self.policy != Policy::Normal);
```
mihirn: Shouldn't this be a `==` instead of a `!=`? Also, do we want to assert here or make it a no-op early return with a warning message?

brayniac: Yes, but I'll just remove it. We already exit early during config validation, so it's really not necessary

PR #568, opened 2025-06-27
feat(agent): add cgroup scheduler stats

brayniac:
Adds scheduler runqueue waiting time as a metric on a per-cgroup and per-CPU basis.

Adds involuntary context switch and offcpu time on a per-cgroup basis.

### DIFF
--- a/src/agent/samplers/scheduler/linux/runqueue/mod.bpf.c
+++ b/src/agent/samplers/scheduler/linux/runqueue/mod.bpf.c
@@ -10,6 +10,7 @@
 // to calculate the runqueue latency, running time, and off-cpu time.
 
 #include <vmlinux.h>
+#include "../../../agent/bpf/cgroup_info.h"
 #include "../../../agent/bpf/helpers.h"
 #include <bpf/bpf_core_read.h>
 #include <bpf/bpf_helpers.h>
@@ -19,11 +20,14 @@
 #define HISTOGRAM_POWER 3
 #define MAX_CPUS 1024
 #define MAX_PID 4194304
+#define MAX_CGROUPS 4096
+#define RINGBUF_CAPACITY 262144
 
 #define TASK_RUNNING 0
 
 // counter positions
 #define IVCSW 0
+#define RUNQ_WAIT 1
 
 /**
  * commit 2f064a59a1 ("sched: Change task_struct::state") changes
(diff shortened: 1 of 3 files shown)


src/agent/samplers/scheduler/linux/runqueue/mod.rs
```diff
@@ -24,13 +24,66 @@ use crate::agent::*;
 
 use std::sync::Arc;
 
+unsafe impl plain::Plain for bpf::types::cgroup_info {}
+
+fn handle_event(data: &[u8]) -> i32 {
```
mihirn, line 29: Out of curiosity, if this is happening per-cgroup, per-sample, is all the string handling a performance concern (and one of the reasons for some of the slow handlers)?

src/agent/samplers/scheduler/linux/runqueue/mod.rs
```diff
@@ -40,10 +93,14 @@ fn init(config: Arc<Config>) -> SamplerResult {
         },
         ModSkelBuilder::default,
     )
-    .counters("counters", counters)
+    .cpu_counters("counters", counters)
```
mihirn, line 96: Curious why this isn't a `packed_counters` given that both the counters are `CounterGroups`

brayniac: packed_counters are tightly packed - eg, each cgroup is sequential without padding. Normal counters which we use for per-cpu have padding to prevent false-sharing across cores.

src/agent/samplers/scheduler/linux/runqueue/mod.rs
brayniac: It's not per-cgroup per sampler - it's per new cgroup. We detect if the serial number and start time mismatch and only read and push info into the ringbuffer when we need to

src/agent/samplers/scheduler/linux/runqueue/mod.rs
mihirn: Got it, thanks.

src/agent/samplers/scheduler/linux/runqueue/mod.rs
mihirn: Ah, that's much less of a worry then.

src/agent/samplers/scheduler/linux/runqueue/mod.bpf.c
```diff
@@ -193,12 +296,57 @@ int handle__sched_switch(u64 *ctx)
...
 	// mark off-cpu at
 	bpf_map_update_elem(&offcpu_at, &pid, &ts, 0);
-	
+
 	// next task has moved into running
 	// - update next->pid running_at with now
 	// - calculate how long next task was enqueued, update hist
 	pid = next->pid;
 
+	// read the next task cgroup details and push to ringbuf if new cgroup
```
mihirn, line 305: Seems like most of this code is common with 217-260 and could be abstracted away, at least if not entirely, into functional bits like zeroing out the counters, reading the name, etc.

brayniac: To be handled across all eBPF samplers in a follow-up PR.

PR #586, opened 2025-07-28
Adding the Docker and Kubernetes deployments.

samof76:
Problem

The rezolus lacked a multi-arch Dockerfile and Kubernetes deployments.

Solution

The following chanages
- Dockerfile: For building the docker images for amd64 and arm64 architectures
- xtask: To have cargo task for build docker images
- Kubernetes deployments: To deploy to kubernetes
- Grafana dashboards: A modeled around the rezolus viewer

Result

The capability to deploy to kubernetes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 !.cargo/config.toml
 target
 **/*.rs.bk
+**/overlays
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,65 @@
+FROM --platform=$BUILDPLATFORM rust:1 AS builder
+
+# Install cross-compilation tools and libraries for both architectures
+RUN dpkg --add-architecture arm64 && \
+    apt-get update && apt-get install -y \
+    clang \
+    build-essential \
+    gcc-aarch64-linux-gnu \
+    libelf-dev \
+    zlib1g-dev \
+    libbpf-dev \
+    libelf-dev:arm64 \
+    zlib1g-dev:arm64 \
+    libbpf-dev:arm64 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Set up working directory
+WORKDIR /app
+
+# Copy Cargo files
+COPY . .
+
+RUN rustup target add x86_64-unknown-linux-gnu && \
+    rustup target add aarch64-unknown-linux-gnu && \
+    rustup component add rustfmt
+
+# Define build argument for target platform
+ARG TARGETPLATFORM
+
+# Set up cross-compilation environment
+RUN case "$TARGETPLATFORM" in \
+    "linux/amd64") \
+    export CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=x86_64-linux-gnu-gcc &&\
+    export PKG_CONFIG_PATH=/usr/lib/x86_64-linux-gnu/pkgconfig &&\
+    cargo build --release --target x86_64-unknown-linux-gnu && \
+    cp target/x86_64-unknown-linux-gnu/release/rezolus /app/rezolus \
+    ;; \
+    "linux/arm64") \
+    export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc &&\
+    export PKG_CONFIG_PATH=/usr/lib/aarch64-linux-gnu/pkgconfig &&\
+    cargo build --release --target aarch64-unknown-linux-gnu && \
+    cp target/aarch64-unknown-linux-gnu/release/rezolus /app/rezolus \
+    ;; \
+    esac
+
+# Final stage: create minimal runtime image
+FROM debian:bookworm-slim
+
+# Install runtime dependencies
+RUN dpkg --add-architecture arm64 && \
+    apt-get update && apt-get install -y \
+    ca-certificates \
+    libelf1 \
+    zlib1g \
+    libbpf-dev \
+    libelf1:arm64 \
+    zlib1g:arm64 \
+    libbpf-dev:arm64 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Copy the built binary
+COPY --from=builder /app/rezolus /usr/local/bin/rezolus
+
+# Set the entrypoint
+ENTRYPOINT ["/usr/local/bin/rezolus"]
--- a/README.md
+++ b/README.md
@@ -194,6 +194,28 @@ target/release/rezolus record http://localhost:4241 rezolus.parquet
 target/release/rezolus hindsight config/hindsight.toml
 ```
 
+### Build with Docker
+
+You can build Docker images using the custom cargo command:
+
+```bash
+# Build for both linux/amd64 and linux/arm64 platforms (requires Docker buildx)
+cargo xtask docker
+
+# Build for a specific platform
+cargo xtask docker linux/amd64
+cargo xtask docker linux/arm64
+
+# Build with a custom tag
+cargo xtask docker both rezolus:v5.2.2
+cargo xtask docker linux/amd64 rezolus:dev
+
+# Or build directly with Docker
+docker build -t rezolus:latest .
+```
+
+The multi-platform build requires Docker buildx and will push to the registry. For local development, use single-platform builds.
+
 ## Contributing
 To contribute to Rezolus first check if there are any open pull requests or
 issues related to the bugfix or feature you wish to contribute. If there is not,
--- a/deploy/kustomize/base/configs/agent.toml
+++ b/deploy/kustomize/base/configs/agent.toml
@@ -1,0 +1,120 @@
+[general]
+listen = "0.0.0.0:4241"
+
+# Metric snapshots are cached with the following TTL to prevent excessive
+# resource usage.
+ttl = "10ms"
+
+[scheduler]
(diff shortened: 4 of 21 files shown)


samof76:
@brayniac Apologies.

1. Grafana dashboard: I will remove them from here, and I guess these can be published seperate through community grafana dashboards
2. Documentation and Rational: I will add a `README.md` to the `deploy` folder, and explaining how to deploy and use rezolus in kuberenetes.
3. Redundant configs: Will remove them from the this, and make it part of the documenation as a dependency.
4. Docker: I would really want you hear your comments on the Dockerfile and `xtask` changes. Also wanted to know why there isn't a `Dockerfile` already?

Will await your response on #591 before proceeding.

README.md
```diff
@@ -194,6 +194,28 @@ target/release/rezolus record http://localhost:4241 rezolus.parquet
 target/release/rezolus hindsight config/hindsight.toml
 ```
 
+### Build with Docker
+
+You can build Docker images using the custom cargo command:
```
brayniac, line 199: Why is this being done with xtask instead of maybe being something we do in CI as part of a publishing workflow? Are users expected to build the docker images themselves? Is this meant for maintainers at publish time? I'm unclear on this.

deploy/kustomize/base/configs/agent.toml
```diff
@@ -0,0 +1,120 @@
+[general]
```
brayniac, line 1: Unclear why we can't just use the configs from `config/` directory. Why is this duplicated?

deploy/kustomize/base/configs/exporter.toml
```diff
@@ -0,0 +1,41 @@
+[general]
```
brayniac, line 1: Same. Not sure why this is a duplicate of config/exporter.toml

deploy/kustomize/base/ds-rezolus.yaml
```diff
@@ -0,0 +1,129 @@
...
+            privileged: true
+            capabilities:
+              add:
+                - SYS_ADMIN
+                - SYS_PTRACE
+                - NET_ADMIN
+                - SYS_RESOURCE
+                - IPC_LOCK
+                - NET_RAW
+          volumeMounts:
```
brayniac, line 32: Are all of these necessary?

deploy/kustomize/base/ds-rezolus.yaml
```diff
@@ -0,0 +1,129 @@
...
+      hostPID: true
+      containers:
+        - name: rezolus-agent
+          image: rezolus
+          args:
+            - /config/agent.toml
+          securityContext:
+            privileged: true
+            capabilities:
+              add:
```
brayniac, line 25: I'm not sure all these capabilities are necessary?

deploy/kustomize/base/kustomization.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - name: rez-exporter-config
+    files:
+      - configs/exporter.toml
+  - name: rez-agent-config
+    files:
+      - configs/agent.toml
+images:
+  - name: rezolus
+    newName: rezolus
+    newTag: v5.2.2
```
brayniac, line 21: Does this get updated with each release? This feels like it could get easily missed

Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM --platform=$BUILDPLATFORM rust:1 AS builder
```
brayniac, line 1: I don't understand why this dockerfile duplicates the build logic. We already build DEB/RPM packages for both x86_64 and arm64 (aarch64) targets. Why the separate build workflow to maintain?

.gitignore
```diff
@@ -1,4 +1,5 @@
 .cargo
 !.cargo/config.toml
 target
-**/*.rs.bk
\ No newline at end of file
+**/*.rs.bk
+**/overlays
```
brayniac, line 5: Why?

README.md
samof76: I wanted to keep it as part of the Cargo's command tooling, which is also a release tool in essence we could keep. I do not really have any strong opinions on this, though.

deploy/kustomize/base/ds-rezolus.yaml
samof76: Yes. Otherwise it will would not pick of the metrics. I am finding it difficult to get the CPU/TLB/IPNS metrics, will mark this PR in WIP, until I freeze on the capabilities. BTW most of my testing is on EKS with AmazonLinux2.

deploy/kustomize/base/ds-rezolus.yaml
samof76: Yes. These are needed otherwise the rezolus was not giving some of the metrics.

.gitignore
samof76: Overlays would specific to the user's use case.

deploy/kustomize/base/configs/agent.toml
samof76: Yes we could do that. The challenge is that if the user needs to override some pieces of it.

Dockerfile
samof76: This is not a seperate workflow, but rather a way to build for multiarch images using buildx. I could not find a better way of doing this.

samof76:
> 1 - I think that just moving the dashboards out still creates some maintenance issue? Who will maintain the dashboards and keep them in-sync with future changes and additions in Rezolus? Instead I wonder if we need not just a separate repo for the dashboards, but some process for generating them. It just seems unreasonable to add more lines of dashboard to the repo than we have code. There must be some better way to handle that?

I think I am aligned there to maintain a seperate repo for grafana dashboards. My bad I should not have included them here. Will remove them.

> 2 - Not just for users, but I need some explanation as to why this is being added in the way it is. This changes adds a lot of seemingly random stuff and gives me no hints or explanation. Decomposing the work into smaller pieces and discussing such large contributions in issues prior to PRs can help provide maintainers clarity and give a chance to weigh-in earlier.

Ok! The primary goal of this PR is to provide the necessary artifacts to do kubernetes deployments. And since we will be deploying to kubernetes, the necessary pre-cursor to that is to have a docker image, hence the Dockerfile and xtask to build docker image, inline with the codebase here.

> 3 - I think we should just use the configs from the existing configs folder? I'm not sure why they got duplicated or would be added to the documentation as a dependency when we already have default configurations in a folder? They get used by our DEB and RPM packaging, so they should be what we use for any container images we may choose to provide.

Yeah I guess thats a good point, we could add the the configs into the docker image itself. But if I am deploying to K8s, there option to provide for customizing configurations is a key. So I kept it the config folder as you see it. But I do agree about the duplication, would rectify it if we have consensus on the whether we need this PR at all or not.

> 4 - There's been no demand for docker images or k8s deployment to date. We have existing DEB and RPM packages that have worked well for deployment on the host level (eg, via configuration management - puppet, ansible, etc)

Ok, I agree, with the till now part. But do you think you will come across cases where that will not hold true in the future. I have no qualms in closing this PR, keeping it docker image and deployment YAMLs in my fork.

README.md
brayniac: Who is intended to run it though?

deploy/kustomize/base/configs/agent.toml
brayniac: So the intent is the configs are edited in-place and the docker containers created and deployed from the repo? I'm unclear what the usage pattern is that this supports.

deploy/kustomize/base/ds-rezolus.yaml
brayniac: I'm surprised it needs NET_RAW.

brayniac: I'm surprised, I don't think we open /dev for anything.

deploy/kustomize/base/ds-rezolus.yaml
brayniac: NET_RAW seems like it shouldn't be needed. CAP_SYS_ADMIN should cover all the BPF progs. You might find that not all of the perf based telemetry works on all instance types in AWS.

Dockerfile
brayniac: This is a list of dependencies and build requirements, isn't it? Can the docker leverage the DEBs we already provide? I don't like the idea of maintaining the list of build dependencies in the Dockerfile in addition to the package control. It feels like this could be done differently.

.gitignore
brayniac: It seems odd to support editing files in the repo? That's a much different workflow than I'm accustomed to.

brayniac:
I think there's something about the general approach in this PR that doesn't feel right. Maybe I'm not familiar with how folks deploying on k8s normally work, but it seems odd to have the user modify files in the repo to do their deployment? The duplication of configuration and build requirements seems wrong and likely to result in drift. The dashboards feel like we should be automatically generating the json instead of editing it in a grafana UI and commiting the exported json.

When looking at all that combined with the lack of documentation and any attempt to add to CI, I'm thinking that it's probably best if you maintain your dockerfile and dashboards separately since that seems to be working for you. Maybe once we have some additional users with requirements for k8s and containerized deployments we will re-evaluate and add this kind of support to our roadmap.

Thanks again for opening this, but I'm going to close it out for now.

Issue #579, opened 2025-07-03
recorder: status message when doing parquet conversion


brayniac:
It would be helpful to have a status message for when we switch from recording to parquet conversion phase.

PR #619, opened 2025-08-27
fix(recorder): add message when converting recording to parquet

brayniac:
Adds an info level log message before conversion to parquet. This provides feedback on when the recording has completed and the recording is being converted / finalized.


brayniac:
This fixes #579

PR #698, opened 2026-03-09
add metric description tooltips to viewer charts

yurivish:
adds the first step towards per-plot metric description tooltips (using the browser 'title' attribute).

extracts common description-related functionality between the mcp server and viewer, then uses them in `PlotOpts`.

Claude's description:

Extract metric_descriptions() into src/common/mod.rs for reuse, replacing the private copy in MCP. In the viewer, attach descriptions to plot options and render an info icon on each chart that shows the metric description on hover via a native title tooltip.


brayniac (changes requested):
rustfmt needed

PR #834, opened 2026-04-27
perf(viewer): defer chart fetches, cap parallelism, stream renders

thinkingfish:
## Summary
Section navigation used to fire one PromQL query per plot up front through `Promise.allSettled`, awaiting all completions before painting. Big sections (CPU, scheduler, syscall) blocked first paint behind 20+ queries. This reworks the fetch pipeline along three axes:

1. **Lazy:** charts fetch their PromQL only when scrolled into view (200px head-start via `IntersectionObserver`).
2. **Capped:** every fetch path (baseline plot, compare experiment, heatmap) routes through a shared `QueryPool` with a cap of 8.
3. **Streaming:** each chart paints as its query lands — no section-wide await barrier.

## Mechanics
- New `QueryPool` (configurable cap, default 8). `setQueryConcurrencyCap(n)` exported for future tuning.
- `processDashboardData` is now a build-only pass: walks plots, runs the existing `buildEffectiveQuery` transform, attaches `_effectiveQuery`, resets `_fetched` flags and bumps a generation token.
- New `fetchPlotData(plot)`: idempotent per generation, queued through the pool. Stale-generation completions short-circuit so a rapid granularity toggle can't clobber the latest result.
- New `LazyChart` Mithril wrapper: `IntersectionObserver` triggers `fetchPlotData` on first intersect (or immediately if already on screen at mount). On redraw, re-arms if `_fetched` was reset.
- `CompareChartWrapper`: experiment fetch moved from `oninit` to `oncreate` + observer, so off-screen compare wrappers stay pending until visible. The view-time refetch on step divergence (PR #833) is preserved.
- `Group` shows a section while plots are pending OR have data; collapses only after fetches settle and none returned data (matching the prior "GPU on a CPU box" hide).
- `flushAll(groups)` exported for any future caller that needs full population (save-with-selection currently doesn't — serializes only chart id + query).

## Test plan
- [x] `node --test tests/*.mjs`
- [x] `cargo check --workspace`
- [x] `./crates/viewer/build.sh`
- [ ] Manual: load a big section (`/cpu` or `/scheduler`) — first row of charts paints quickly; subsequent rows fill in as you scroll
- [ ] Manual: granularity toggle re-fetches all visible charts; off-screen ones repopulate when scrolled to
- [ ] Manual: compare mode — same visible-first behavior on both baseline and experiment series; the existing step-diverge refetch still updates already-fetched compare charts
- [ ] Manual: heatmap toggle — fanouts route through the pool (concurrency bounded)
- [ ] Manual: section that returns no data on this host (e.g. GPU-only section on a CPU-only box) collapses after fetches settle

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2794,7 +2794,7 @@ dependencies = [
 
 [[package]]
 name = "rezolus"
-version = "5.11.1-alpha.10"
+version = "5.11.1-alpha.11"
 dependencies = [
  "allan",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ wasm-bindgen = "0.2"
 
 [package]
 name = "rezolus"
-version = "5.11.1-alpha.10"
+version = "5.11.1-alpha.11"
 description = "High resolution systems performance telemetry agent"
 edition = "2021"
 license.workspace = true
--- a/src/viewer/assets/lib/data.js
+++ b/src/viewer/assets/lib/data.js
@@ -12,6 +12,59 @@ const setStepOverride = (step) => { _stepOverride = step; };
 const getStepOverride = () => _stepOverride;
 
 // ---------------------------------------------------------------------------
+// Concurrency-capped query pool. Shared by every fetch path (baseline
+// per-plot, compare-mode experiment, heatmap) so we never have more than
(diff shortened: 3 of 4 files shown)


thinkingfish:
Closing — broke the standalone viewer. Reverting.

PR #855, opened 2026-04-30
perf(deps): bump metriken-query 0.9.5 -> 0.9.6 + memory regression test

thinkingfish:
## Summary

Bump metriken-query 0.9.5 → 0.9.6, plus an alpha rev bump to 17.

Upstream 0.9.6 substantially reduces peak heap during `Tsdb::load_from_bytes` (the eager-decode path the WASM viewer takes on attach).

## Memory savings — measured

Peak heap allocations during `Tsdb::load_from_bytes` for the three representative demo parquets:

| File | 0.9.5 | 0.9.6 | Savings |
|---|---|---|---|
| `demo.parquet` | 88 MiB | 9 MiB | **−90%** |
| `cachecannon.parquet` | 239 MiB | 56 MiB | **−77%** |
| `disagg/sglang-nixl-16c.parquet` | OOM-prone in WASM | 171 MiB | comfortably under wasm32's 4 GB cap |

The nixl combined parquet (5 sources, ~5000 columns) was hitting `handle_alloc_error` in the browser pre-upgrade — the panic surfaced as "Failed to load demo: unreachable" we diagnosed earlier. With 0.9.6 it fits comfortably.

## Methodology

Numbers above were produced with a one-off integration test (not shipped) that wraps the system allocator with a `GlobalAlloc` tracking peak resident bytes — deterministic, immune to OS RSS reclaim quirks. Each fixture was loaded via `Tsdb::load_from_bytes`, peak measured, dropped, repeat. Test run in `--release` (debug-mode allocator behavior is too noisy).

## Tests
- [x] `cargo build --bin rezolus` — clean
- [x] `cargo clippy --bin rezolus -- -D warnings` — clean
- [x] `cargo test --bin rezolus` — 105 passed

## Test plan
- [x] Manual: load `disagg/sglang-nixl-16c.parquet` in the WASM viewer (after a `crates/viewer/build.sh` rebuild) and verify it loads without the previous OOM panic


thinkingfish:
Will be merged into #856

PR #857, opened 2026-05-01
fix(dashboard): per-host disambiguation in cpu/overview/softirq queries

thinkingfish:
## Summary

The CPU/Overview/Softirq dashboards were written assuming single-host data. On a multi-host (combined-rezolus) parquet, queries shaped like

```promql
sum(irate(cpu_usage[5m])) / cpu_cores / 1000000000
```

produce a blank panel because:

- LHS `sum(...)` collapses to `{}`
- RHS `cpu_cores` carries `{node=…}` per host (one series per host in a combined file)
- the implicit vector-match between `{}` and `{node=…}` fails → empty result

Rewrite to per-host:

| before | after |
|---|---|
| `sum(...)` | `sum by (node) (...)` |
| `sum by (id) (...)` | `sum by (id, node) (...)` |

Each chart now renders one line per host on a multi-host file. On single-host data both forms degenerate to today's behavior — series without a `node` label all fall into the `{}` group, producing the same single line as before.

## Files touched

- `crates/dashboard/src/dashboard/cpu.rs`
- `crates/dashboard/src/dashboard/overview.rs`
- `crates/dashboard/src/dashboard/softirq.rs`

## Skipped: throughput-efficiency queries in overview

The "Normalized by Throughput" group at the bottom of `overview.rs` divides fleet CPU/network by a service throughput KPI. Those intentionally collapse the LHS to `{}` so it can join against the throughput KPI's `{instance=…}` labels. Per-host disambiguation would break that flow. Left untouched with an inline comment explaining why.

## Tests
- [x] `cargo build --bin rezolus` clean
- [x] `cargo clippy --bin rezolus -- -D warnings` clean
- [x] `cargo test -p dashboard` clean
- [x] Sanity-check on `site/viewer/data/demo.parquet` (single-host): rewritten query returns identical 301-point series, mean 0.996 — same data, same values
- [ ] Manual: load a combined multi-host rezolus parquet (e.g. one of the disagg fixtures) and verify CPU/Overview/Softirq sections now render per-host lines

### DIFF
--- a/crates/dashboard/src/dashboard/cpu.rs
+++ b/crates/dashboard/src/dashboard/cpu.rs
@@ -14,12 +14,12 @@ pub fn generate(data: &Tsdb, sections: Vec<Section>) -> View {
     busy.describe("Overall CPU busy time across all cores, with per-core breakdown.");
     busy.plot_promql(
         PlotOpts::counter("Busy %", "busy-pct", Unit::Percentage).percentage_range(),
-        "sum by (node) (irate(cpu_usage[5m])) / cpu_cores / 1000000000".to_string(),
+        "sum(irate(cpu_usage[5m])) / cpu_cores / 1000000000".to_string(),
     );
     busy.plot_promql(
         PlotOpts::counter("Busy % (Per-CPU)", "busy-pct-per-cpu", Unit::Percentage)
             .percentage_range(),
-        "sum by (id, node) (irate(cpu_usage[5m])) / 1000000000".to_string(),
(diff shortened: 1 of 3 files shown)


thinkingfish:
Closing — wrong diagnosis. The frontend already injects `{node="X"}` into every non-service section query via `injectLabel` (data.js:387-388) when a node is selected. The original `sum(irate(metric[5m]))/cpu_cores` queries are correct; they get scoped to the selected node by the frontend. My rewrite to `sum by (node) (...)` would have produced N lines per chart on multi-host files instead of the intended single line for the selected node. Reverted on the branch; net diff is zero.

PR #873, opened 2026-05-06
feat: Exceptions dashboard + blockio errors/requeues sampler

thinkingfish:
## Summary

Centralized **Exceptions** view that aggregates terminal fault / dropped-work metrics across subsystems, plus a new BPF sampler for the missing piece (block IO errors and requeues), with a companion diagnostic-patterns doc. Iterated polish in subsequent commits to pull the section into its own nav slot, unify the chart shapes, and fix a few palette / layout issues that surfaced during testing.

### 1. Exceptions dashboard section (\`crates/dashboard/src/dashboard/exceptions.rs\`)

Single screen for an operator to see whether the host is dropping packets, getting throttled, retransmitting, hitting cloud-imposed rate limits, or failing IO. Five groups, all driven by metrics already available (or added in #2 below):

| Group | Metrics |
|---|---|
| **Network** | \`network_drop\`, \`network_transmit_timeout\` |
| **TCP** | \`tcp_retransmit\` |
| **AWS ENA Allowance Exceeded** | \`network_ena_*_allowance_exceeded\` (bandwidth rx/tx, PPS, conntrack, link-local) |
| **CPU Throttling** | \`cgroup_cpu_throttled\`, \`cgroup_cpu_bandwidth_throttled_periods\`, throttled-time variants |
| **Block IO** | \`blockio_errors\` and \`blockio_requeues\` from #2, rendered as multi-line full-width charts (By Class / By Op for errors, By Op for requeues) |

Section name was updated from "Errors" to "Exceptions" to match the operator-facing copy elsewhere; metric names (\`blockio_errors\`, \`network_drop\`, etc.) are unchanged because those are accurate at the kernel/protocol level. Empty groups render naturally for hosts that lack the underlying metrics (non-AWS for ENA, non-eBPF for blockio_errors), so the section is self-gating.

**Section placement:** Exceptions sits right after Overview in \`SECTION_META\` and renders as a top-level sidebar entry above the "Samplers" label — it's a cross-sampler triage view, not a per-sampler section.

### 2. New BPF sampler \`blockio_requests\` extensions (\`src/agent/samplers/blockio/linux/requests/\`)

Block IO failures weren't being captured by any existing sampler. The errors + requeues counters were folded into the existing \`blockio_requests\` sampler in a follow-up commit (was originally a separate \`blockio_errors\` sampler). Two BPF tracepoints:

- **\`block:block_rq_complete\`** — early-return on \`BLK_STS_OK\` so the hot path is one comparison; classify non-OK status into one of seven coarse buckets (\`io\`, \`timeout\`, \`nospc\`, \`target\`, \`protection\`, \`unsupported\`, \`other\`).
- **\`block:block_rq_requeue\`** — counts requests put back on the queue (SCSI EH escalation, NVMe controller reset, multipath path failover) — recovered events distinct from terminal errors.

Two metrics, both labeled by op (\`read\` / \`write\` / \`flush\` / \`discard\`):

- \`blockio_errors{op, error}\` — 28 series (4 ops × 7 error classes)
- \`blockio_requeues{op}\` — 4 series

Per-CPU bank widths in the BPF maps are padded to whole 64-byte cachelines (32 slots for errors, 8 for requeues) to match the userspace counter reader's layout — no contention or false sharing on the increment path.

The error-class bucketing is operationally driven: pairing \`blockio_errors\` and \`blockio_requeues\` makes it possible to distinguish recovered transport blips (high requeues, flat errors) from real damage (both rising) at a glance. See \`docs/patterns.md\` for the full interpretation guide.

### 3. Diagnostic patterns doc (\`docs/patterns.md\`)

Eight diagnostic patterns covering how to read metric shapes to identify recurring failure modes: block IO error mix interpretation (table form), IRQ pinning misconfiguration, cloud VM contention vs self-induced overload via steal-time decomposition, CPU throttling vs scheduling pressure, ENA allowance limits, TCP retransmits, live-migration window detection, and off-CPU time decomposition.

## Polish along the way

- **Multi-line chart shapes.** Block IO errors initially rendered as 7 single-line charts (one per error class) + 4 (one per op). Collapsed to 2 multi-line charts using \`sum by (error)\` and \`sum by (op)\`, mirroring the existing Requeues pattern. Total charts dropped — multi-line charts already show breakdown + relative magnitudes.
- **Full-width by default for multi-line.** The collapsed charts use \`plot_promql_full\` so the legend has room to breathe.
- **Per-chart descriptions.** Split the long Block IO subgroup blurb into per-chart subtitles via a new \`PlotOpts::with_description\` builder. Description renders as a block above the chart card (not a subtitle inside the absolute header overlay), at body font size for readability.
- **Color palette.** Multi-line charts pull colors from the shared \`ColorMapper\` (hash-by-name). Swapped from Tailwind 500 → d3 \`schemeCategory20\`, then dropped the muted brown + gray pairs so hash-indexed series never roll the dice on a beige slot. \`CHART_PALETTE\` (default echarts \`color\` for series-index-mapped charts) is d3 \`schemeCategory10\`.
- **cgroup selector layout fix.** Drive-by — the two-column cgroup picker had lost equal-width sizing because the name span was missing \`min-width: 0\` in its flex chain. Fixed.

## Wiring

- \`build.rs\` — added the new BPF sources to \`SOURCES\`.
- \`src/agent/samplers/blockio/linux/mod.rs\` — module registration.
- \`config/agent.toml\` — sampler enable stanza.

## Test plan

- [x] \`cargo build --bin rezolus\` clean
- [x] \`cargo build -p dashboard\` clean
- [x] \`cargo clippy --all-targets -- -D warnings\` clean
- [x] \`cargo test --bin rezolus\` (116/116 pass)
- [x] \`cargo run -p dashboard -- /tmp/out\` — \`exceptions.json\` is produced and the **Exceptions** entry appears in \`sections.json\` between Overview and the per-sampler list
- [x] \`bash tests/viewer_smoke.sh\` — viewer smoke (PR #893) passes against this branch
- [x] Visual verification on \`/exceptions\`: Block IO renders 2 multi-line charts under Errors (By Class, By Op) + 1 multi-line under Requeues (By Op), all full-width, descriptions float above each chart at body font size, no beige in the rotation
- [x] Visual verification on \`/cgroups\`: column widths equal again, names ellipsize on overflow
- [x] Run the agent on a Linux host with active block IO and verify \`blockio_errors\` / \`blockio_requeues\` populate as expected (deferred — needs a real Linux host with eBPF and a fault-injection workload; sandbox doesn't have eBPF)


docs/TODOs.md
```diff
@@ -0,0 +1,72 @@
+# Suggested Additions to Rezolus
+
+Items raised during the Exceptions-dashboard work that weren't in scope for that PR but would close visibility gaps in adjacent areas.
+
+## Samplers
+
+### Hardirq accounting
+
+Today rezolus tracks per-CPU `softirq` time and counts but not hardirq. `/proc/interrupts` exposes a per-CPU count for every IRQ vector (`LOC`, `RES`, `CAL`, `TLB`, `IWI`, `MCE`, plus per-device IRQs). A new procfs sampler that diffs these counts per scrape interval would directly show:
```
brayniac: I don't think I agree with doing it this way.

docs/TODOs.md
```diff
@@ -0,0 +1,72 @@
+# Suggested Additions to Rezolus
```
brayniac: Overall, I don't think these are fully thought-out. If it's currently lacking in the CLAUDE.md or other doc sources, we should try to spell out some stronger guidelines about the approach we're taking overall and why. It seems that Claude doesn't quite understand the guiding principles - for example: avoid bpf hashmaps, try to use BPF instead of procfs or sysfs reads, etc. I feel like our intentions in the approach are what make Rezolus special. 

I'd shelf this and drop it from the PR. There's some interesting ideas here, but I'd rather extract the essence of what each suggestion is and file issues about what could be instrumented and what value it would bring. Separating that from a prescribed approach gives us cleaner feature requests.

src/agent/samplers/blockio/linux/errors/mod.bpf.c
```diff
@@ -0,0 +1,148 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2024 The Rezolus Authors
```
brayniac: Wrong year.

src/agent/samplers/blockio/linux/errors/mod.bpf.c
```diff
@@ -0,0 +1,148 @@
...
+    case BLK_STS_PROTECTION:
+        return 4; // protection
+    case BLK_STS_NOTSUPP:
+        return 5; // unsupported
+    default:
+        return 6; // other
+    }
+}
+
+SEC("raw_tp/block_rq_complete")
```
brayniac: This adds another probe at a tracepoint we instrument for both blockio latency and request samplers. The entry/exit cost are higher than what we're doing in the probe. I'd prefer to consolidate and would consider merging this into blockio_requests sampler instead. Folding the requests and latency ones together after that would bring additional benefit. But I'd rather try to avoid attaching another BPF program here.

brayniac:
I would discourage merging as-is.

EDIT: The recent changes addressed my concerns.

docs/TODOs.md
thinkingfish: Agree. These are not at all thought-out. I summarized some interesting feature ideas when carrying out some other code changes so we don't completely lose track of them. I have re-written the doc to be why & what only, without the Hows. The plan was always to go through each and decide whether and how to implement them much more carefully.

src/agent/samplers/blockio/linux/errors/mod.bpf.c
thinkingfish: Good point. This is now rewritten and merged into the existing ebpf program.

Overall I think it'll be great to have a skill or guiding doc about the tradeoff and methodology of Rezolus sampler dev, as you suggested in another comment.

brayniac (approved):

PR #897, opened 2026-05-10
Use cargo install for wasm-pack instead of shell script

thinkingfish:
## Problem

The current release workflow installs `wasm-pack` by downloading and executing a shell script from the internet. This approach has several drawbacks:
- It's less reproducible and harder to pin to a specific version
- Shell script execution can be slower and less reliable
- It doesn't integrate well with Rust tooling conventions

## Solution

Replace the shell script installation method with `cargo install wasm-pack --version ^0.14 --locked`, which:
- Uses Cargo's standard package installation mechanism
- Pins the version to `^0.14` (compatible with 0.14.x releases)
- Uses `--locked` to ensure reproducible builds with locked dependency versions
- Integrates better with the existing Rust toolchain

## Result

The `wasm-pack` installation in the release workflow will be more reliable, reproducible, and maintainable while maintaining compatibility with the 0.14.x version series.

https://claude.ai/code/session_01QD42w778EpqJ3Fuz13j79y

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -240,7 +240,7 @@ jobs:
           targets: wasm32-unknown-unknown
 
       - name: install wasm-pack
-        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+        run: cargo install wasm-pack --version ^0.14 --locked
 
       - name: build wasm viewer
         run: ./crates/viewer/build.sh


thinkingfish:
Superseded by #898. The cherry-picked commit on this branch is identical to #896 (already merged to main), so rebasing this onto main leaves an empty diff. #898 takes the cleaner approach: revert the post-release alpha bump on top of main so the v5.13.0 tag can sit on a commit that's actually reachable from `main`. Closing.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #934, opened 2026-05-16
feat(viewer): <rezolus-chart> embed component with exact viewer parity

thinkingfish:
## Summary

A `<rezolus-chart>` web component for embedding a single dashboard chart outside the viewer, plus a `/lib/embed/demo.html` spike page. The component renders **identically to the viewer** by mounting the viewer's real chart pipeline rather than reimplementing it.

- **`<rezolus-chart>`** (Lit, shadow DOM): card chrome from `/lib/charts.css`; the chart itself is the viewer's actual `Chart` mithril component mounted into the shadow root with a throwaway `ChartsState`. Descriptors flow through the same `configureChartByType` dispatcher the dashboard uses, so line/scatter/heatmap/quantile/multi render byte-for-byte (unit-aware axes & tooltip, fonts, gridlines, dataZoom, OOB bands). `COLORS` read CSS tokens from `document.documentElement`; a `data-theme` MutationObserver remounts so a theme flip re-reads them.
- **`demo.html`** is a faithful, self-contained harness:
  - The viewer is query-driven (`/data/rezolus.json` carries descriptors, never inline series), so the demo executes the plot's `promql_query` via `/api/v1/query_range`, mirroring `data.js::executePromQLRangeQuery`'s window/step recipe.
  - Scopes the query to the selected node exactly like the viewer — picks `pinned_node` (else `nodes[0]`) from `/api/v1/file_metadata` and applies the viewer's own exported `injectLabel`, so a multi-node recording doesn't fan out across nodes.
  - Graceful fallbacks: synthetic series only when the query genuinely returns nothing or the fetch fails.
- `demo.html` loads `echarts` + `mithril` (`window.m`) as globals since the reused `Chart`/`chart.js` use them.

Requires running inside (or behind) the rezolus viewer so the absolute `/lib/...`, `/data/...`, `/api/v1/...` paths resolve — a plain static server won't work.

## Test plan

- [x] Endpoint shapes verified live against `cachecannon.parquet`: `/api/v1/metadata`, `/data/rezolus.json` (descriptor, empty inline data), `/api/v1/query_range`, `/api/v1/file_metadata` (2 nodes, pinned `rezolus-server`)
- [x] Node scoping confirmed: injected query yields the per-node series (~0.0058) vs the old combined-node sum (~0.014)
- [x] Full ES module graph (`rezolus-chart` → `charts/chart.js` → deps, `lit`, `data.js`) resolves 200 over `/lib/`
- [ ] Manual: open `/lib/embed/demo.html` next to the live viewer CPU chart — green status, pixel parity, theme toggle, fallback states

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2756,7 +2756,7 @@ dependencies = [
 
 [[package]]
 name = "rezolus"
-version = "5.13.1-alpha.9"
+version = "5.13.1-alpha.10"
 dependencies = [
  "allan",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ wasm-bindgen = "0.2.120"
 
 [package]
 name = "rezolus"
-version = "5.13.1-alpha.9"
+version = "5.13.1-alpha.10"
 description = "High resolution systems performance telemetry agent"
 edition = "2021"
 license.workspace = true
--- a/src/viewer/assets/lib/embed/demo.html
+++ b/src/viewer/assets/lib/embed/demo.html
@@ -53,6 +53,10 @@
     <rezolus-chart id="chart"></rezolus-chart>
 
     <script type="module">
+        // Reuse the viewer's exact label-injection so the embed scopes
+        // the query to one node the same way the dashboard does.
+        import { injectLabel } from '../data.js';
+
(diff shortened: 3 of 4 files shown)


thinkingfish:
Duplicate of #915 (created from fork by mistake). The work continues on #915.

PR #937, opened 2026-05-20
Use DuckDB

yurivish:
# viewer: migrate all query paths to DuckDB SQL; delete metriken-query

## Summary

Retires the in-memory PromQL evaluator and routes every viewer path
(file / upload / A-B / live-agent), MCP, `parquet annotate`, Save-as-Report
column trim, and the `validate_service_extensions` load-time check onto
`metriken_query_sql::DuckDbBackend`. The legacy `metriken-query` crate
(Tsdb + PromQL evaluator + harness; ~16.7K LOC counting everything that left
with it) is deleted; the `live-mode` / `sql-only` feature seam is gone; the
build matrix collapses to a single `cargo build --bin rezolus`.

Why now: PromQL had become the long pole — duplicated query semantics with
the WASM viewer (which already spoke DuckDB), two build configurations, and
a blocker for features that want real SQL surface (`mcp query`, raw-SQL
explorers, server-side `_src_node_*` rewriting). Unifying on DuckDB is what
makes the two viewers structurally identical and the engine swappable.

Pairs with the **companion metriken PR**, which lands `DuckDbBackend`,
`LiveSource`, the H2 histogram UDFs, and `shared_macros.sql`. Both repos must
land together; cross-repo `include_str!` is intentional, documented at
`crates/viewer-sql/src/lib.rs`. 143 commits, 168 files, +16.5K / -6.9K.

Deeper reading lives in `review/`:

- `review/primer.md` — testing primer; the handoff entry point for someone
  who knows Rezolus but not this branch.
- `review/review.md` — full review: architecture, carve-outs, PromQL purge
  narrative, commit-level notes, verification recipe.
- `review/architecture.md` — newcomer tour with ASCII data-flow diagram.

The metriken side has a parallel `review/` with `review.md` (engine review,
concurrency story, deletion narrative) and `architecture.md` (engine
architecture / crate map).

## What changed structurally

- **`Arc<DuckDbBackend>` on `AppState` is the single query engine.** Parquets
  wrap in `SqlCapture` (`sql_capture.rs`); live captures wrap an
  `Arc<LiveSource>` in `LiveCapture` (`live_capture.rs`) and register under
  `LIVE_BASELINE_DATA_SOURCE = "live:baseline"`. `data_source_for` resolves
  live ahead of any parquet path so `/api/v1/query{,_range}` dispatch is
  uniform across modes.
- **`crates/dashboard/src/sql.rs`** — 21 `pub fn` builders (`rate_5m_total`,
  `irate_total`, `hist_percentile_series`, `cpu_pct_total`,
  `cgroup_irate_by_name`, `bucket_heatmap_sql`, `percentile_kpi_sql`, …) that
  ~170 plot call sites funnel through. 25 snapshot tests pin emitter output.
- **`crates/prom-matrix/`** — shared Arrow → Prometheus-matrix JSON projection.
  One envelope formatter, two entry points (`arrow_to_prom_matrix` native,
  `js_arrow_to_prom_matrix` WASM); the JSON shape can't drift.
- **SQL handler runs under `tokio::task::spawn_blocking`** (`7fc2f4d`) so 20+
  parallel chart fetches on first paint don't starve the runtime — worst-case
  query wall time on vllm.parquet dropped from multi-second to ~700 ms.
- **`/api/v1/section_status`** server-driven sidebar gate
  (`d048379` + `f47cbba` + `69ff6b5` + `87a8aae`): empty sections gray out
  before first paint instead of letting users click into voids.
- **`metriken-query` gone from the dep graph.**
  `cargo tree -p rezolus | grep 'metriken-query '` is empty; `promql-parser`
  drops out entirely.

### Identical to before

Every chart on every dashboard. A/B compare, Save-as-Report round-trip, upload,
live-agent, WASM static site, events feature (main's contribution, integrated
through the new SQL save paths). Sparse-metric plots still render as empty
matrices — preserved by a binder-error shim in `run_sql`. End-to-end browser
audit across `demo.parquet` and `cachecannon.parquet` on all 12 pages: every
plot binds, 209/218 KPIs ship SQL (the 9 unavailable all in
`inference-library.json`, a placeholder template, rendering as `_unavailable`
cards).

### Intentionally different

- **`mcp query` takes DuckDB SQL, not PromQL.** Breaking CLI change; M-in-MCP
  clients are LLMs and SQL-fluent. `detect-anomalies` / `analyze-correlation`
  still accept bare metric names (auto-resolved to canonical SQL by kind).
- **First query per parquet pays cold-start** — hundreds of ms while
  `DuckDbBackend` opens the file, registers UDFs + macros, and materializes
  `_src` / `_cgroup_index`. Subsequent queries hit a warm slot; live mode has
  no cold-start.
- **Per-cgroup plots show a ~5-min tail of carried-forward rate past an
  exited cgroup's last sample** (`67e9b95`). Matches PromQL's `[5m]` lookback;
  the SQL pipeline was wrongly truncating before this fix.
- **Sidebar sections gray out** when no plots in the section have data —
  previously every section was live regardless.

## Pre-existing-on-main caveats (NOT regressions of this branch)

Reproducible against `origin/main` on the same parquets — documenting so a
reviewer doesn't suspect the SQL migration. All filed as follow-ups.

- **`/cgroups` empty-state silent-render before the user picks pairs** —
  wants an empty-state callout in `cgroup_selector`.
- **Live-mode Save-as-Report drops events** — `snapshots_to_parquet` skips
  the `KEY_EVENTS` footer entry the parquet path stamps.
- **WASM static-site Save-as-Report non-functional** — calls a WASM binding
  that isn't exported.
- **`COLUMNS()` regex trim footgun** — catalog-text-scan trim can drop
  columns referenced by regex-only queries. Documented at
  `crates/report-save/src/lib.rs:136-139`.

## Acknowledged scope-limit cut

Overview's **Normalized by Throughput** group (`CPU Time / Throughput`,
`Network TX / Throughput`, `Network RX / Throughput`) is not re-emitted in
SQL. Main built these by dividing a sampler metric by the service-extension's
throughput KPI's PromQL; the SQL transcription needs a new accessor on
`ServiceExtension::throughput_query_sql()` plus an emitter pass in
`overview.rs`. Filed as follow-up — on the demo cachecannon parquet these
rendered as `_unavailable` notes on main anyway (PromQL ran but returned no
data).

## Verification

```bash
cargo build --bin rezolus                            # single build matrix
cargo tree -p rezolus | grep 'metriken-query '       # empty
cargo test --bin rezolus                             # 192 pass / 0 fail
cargo test -p dashboard -p prom-matrix -p viewer-sql -p report-save
cargo test --test mcp_cli                            # MCP end-to-end
node --test tests/*.mjs                              # 137 pass / 0 fail
bash tests/viewer_smoke.sh                           # file/upload/A-B/proxy
bash scripts/viewer_chromium_smoke.sh site/viewer/data/cachecannon.parquet
bash scripts/viewer_chromium_smoke.sh --live http://localhost:4241 --ingest-wait 5
```

The chromium per-section smoke is the load-bearing check: it walks every
section in `/api/v1/sections` and asserts each one either rendered a real
chart, reserved an `_unavailable` placeholder, or displayed a `.section-notes`
no-data callout. Catches the silent-render failure mode the API-only smoke
can't see (200 OK with blank body). Requires `chromium`, `jq`, `python3`,
`pip install --user websockets`. Engine-side tests live in the companion
metriken repo (`cargo test -p metriken-query-sql`).

## Reviewer guide — where to spend attention

1. **Carve-outs in `review/review.md`** — the two remaining structural gaps
   (multi-node *set* selection; multi-rezolus aggregation). Single-node now
   filters server-side via `rewrite_src_to_node_view`.
2. **`src/viewer/routes.rs::run_sql`** — the binder-error → empty-matrix shim
   that preserves "unknown metric → empty series". Concentrated complexity.
3. **`crates/dashboard/src/sql.rs`** — the 21 emitters every plot funnels
   through. Snapshot-tested.
4. **`crates/prom-matrix/`** — the projection layer shared between server and
   WASM; the JSON shape contract.
5. **`src/viewer/live_ingest.rs`** + companion `LiveSource` —
   `canonical_column_name` keeps `_src` byte-identical between parquet and
   live so the same dashboard SQL binds in both modes.

The mechanical `Tsdb` → `DuckDbBackend` move is straightforward and has full
end-to-end coverage; the design questions are where to spend attention.

## Companion PR

Pairs with the **companion metriken PR** (lands `DuckDbBackend`, `LiveSource`,
`canonical_column_name`, H2 histogram UDFs, `shared_macros.sql`). Both repos
must land together; see `review/review.md` for the engine-side companion
narrative.


crates/prom-matrix/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "prom-matrix"
+version = "0.1.0"
+edition = "2024"
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+publish = false
+
+# Shared Arrow → Prometheus-matrix JSON projection. Native consumers (the
```
brayniac, line 10: I'm also confused about all the comments in this manifest.

crates/prom-matrix/src/lib.rs
```diff
@@ -0,0 +1,80 @@
+//! Arrow → Prometheus matrix-shape JSON.
```
brayniac, line 1: I'd hope for the rustdoc for the entry to a new crate to explain what it's about in human language. My guess is that this is some translation crate between two different shapes. But it'd be nice to succinctly state what it does and when one would use it.

I guess a higher-level question here is why? Is there any reason to keep the promql shape? The motivation here is unclear

crates/viewer-sql/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "viewer-sql"
+version = "0.1.0"
+edition = "2024"
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+publish = false
+
+# Browser-side viewer that drives duckdb-wasm via JS. Mirrors the
```
brayniac, line 10: Same here as with other places. It'd a little odd to document like this in comments in a Cargo manifest. A short package description is certainly welcome, but typically you want this in the rustdoc at the top of lib.rs instead. 

Same higher-level question about motivation for replacing viewer with viewer-sql. I assume this is another artifact from a iterative replacement, but this should be `viewer` and just bump the version appropriately

PR #992, opened 2026-07-06
feat(drivehealth): per-drive temperature via pass-through ioctls

brayniac:
## Summary
- New Linux `drivehealth` sampler: per-drive temperature for **NVMe and SATA**, read directly from the device via **read-only pass-through ioctls** (the `smartctl`/`hddtemp` mechanism) — **no kernel module** (`drivetemp` not required). SATA uses `SG_IO` ATA PASS-THROUGH (SMART attr 194); NVMe uses admin Get Log Page 0x02.
- Reads are device commands (~7.6 ms/drive), so they're throttled to a configurable `interval` (default 60s) and dispatched off the sample cycle via `spawn_blocking`; `refresh()` stays ~2 µs. Adds a per-sampler `interval` config knob.
- Emits `drive_temperature{device,type,model,serial}`. Unprivileged / no-drive hosts fail closed (zero series, no error). Only read-only diagnostic commands are ever issued.
- Also codifies the review discipline this effort surfaced: `docs/principles.md` principles **16** (measure sampler overhead; ship the number) and **17** (throttle + decouple expensive userspace reads), a new **`reviewing-samplers`** skill, and an `engineering-journal` skill rule. Design/decision history (incl. the hwmon→pass-through pivot) is in `docs/journal/2026-07-06-drive-health-sampler.md`.

## Test plan
- [x] `cargo test` — 227 pass, 3 ignored (hardware), 0 fail
- [x] `cargo clippy` / `cargo fmt --check` clean
- [x] Unit tests: ATA/NVMe temperature parsers (fixture bytes), sysfs enumeration (fixture trees), config `interval` override/default
- [x] **Hardware-verified** on 23 SATA drives (behind a SAS HBA, `drivetemp` NOT loaded): 23/23 read, values match `smartctl -A` per drive; async `refresh()` populates all 23 gauges; measured `refresh()` ~2 µs, full sweep ~176 ms off-cycle; unprivileged → fails closed
- [ ] NVMe path: parser fixture-verified only — **needs validation on a host with an NVMe drive** (documented reopen condition)
- [ ] SAS (true SCSI `LOG SENSE`): deferred — no SAS-only hardware to verify


brayniac:
The main consideration when reviewing this is whether we're okay with this truly async sampling that occurs outside of the collection window. I imagine for thermals this is fine, but it is worth considering carefully. The short version is that instead of blocking the sampler refresh cycle on the completion of the ioctls to read drive temperatures from every drive, we trigger a refresh that will be visible in a future sample. This means the value returned on the metrics endpoint may be stale (beyond just the per-sampler TTL).

thinkingfish:
Given the cadence, I wonder if we should get a timestamp when the reading returns, instead of using timestamp from the sampling loop driver.

brayniac:
> Given the cadence, I wonder if we should get a timestamp when the reading returns, instead of using timestamp from the sampling loop driver.

Yeah. I think there's some things we can do there. I think the easy answer for now is to just pull in temperature as a broad signal, but pull in thermal throttling counters as a more definitive sign. It won't be perfect, but as a health signal we can probably still correlate "throttling happened around this time, and we saw latency spike". 

I think the timestamp per source thing requires more thought.

PR #1012, opened 2026-07-17
fix(viewer): missing boxplot.js symlink (deployed viewer down) + CI guard against recurrence

thinkingfish:
## Symptom

The **deployed static (WASM) viewer doesn't come up at all** — blank page, load-time failure.

## Root cause

`src/viewer/assets/lib/charts/boxplot.js` (from the display-mode/decimation work) is statically imported by `line.js`/`multi.js`/`scatter.js` — core chart renderers loaded at bootstrap — but its `site/viewer/lib/charts/boxplot.js` symlink was never created. The server viewer serves `src/` directly (fine); the static viewer loads `./lib/charts/boxplot.js`, **404s**, the ES-module chain fails, and the app never boots. (`charts/` is a real dir with per-file symlinks, so a new file there needs its own symlink — unlike `embed/`/`lit/`, which are directory symlinks that already cover their files.)

## Fix (commit 1)

Add the one missing per-file symlink `site/viewer/lib/charts/boxplot.js`. One line, no code change.

## Guard against recurrence (commit 2)

The class of bug — a shared `src/viewer/assets/lib` module with no resolving `site/viewer/lib` entry — was only checked by a **local Claude Code pre-commit hook** whose wiring is per-checkout config, so plain-git/CI never ran it and the gap merged. This commit makes it a **binding CI gate**:

- `scripts/check-viewer-symlinks.sh` + `.github/workflows/viewer-symlinks.yml` — fails any PR where a shared viewer module doesn't **resolve** under `site/viewer/lib/`. It's resolution-based (`[ -e ]`), so files covered by a directory symlink (`embed`, `lit`) are correctly treated as present. Verified: it **catches** the boxplot gap (exit 1) and passes on the fixed tree (55 modules).
- The pre-commit hook now calls the same canonical script (hook == CI), fixing its two staleness bugs: a per-file-only `-L` check that false-flagged directory-symlinked files, and a dead `data.js → data_base.js` special case.
- Corrected the `sync-viewer-symlinks` + `viewer-parity` skills: the only standalone files are `script.js` + `viewer_api.js` (`data.js` is a same-name symlink; there is no `data_base.js`; `dashboards.js` is gone); coverage may be a directory symlink; CI is the binding gate.

No embed/lit change needed — they're directory-symlinked and were never missing (a tooling false-positive during triage; corrected).

### DIFF
--- a/.claude/scripts/pre-commit-check.sh
+++ b/.claude/scripts/pre-commit-check.sh
@@ -63,35 +63,15 @@ check_clippy() {
 # ── 3. Check symlinks ───────────────────────────────────────────────
 
 check_symlinks() {
-    # Walk every .js and .css file under src/viewer/assets/lib/
-    while IFS= read -r src_file; do
-        rel="${src_file#$SRC/}"          # e.g. "charts/metric_types.js" or "theme.js"
-        base="$(basename "$rel")"
-        dir="$(dirname "$rel")"          # "." for top-level
-
-        # Skip top-level standalone files
-        if [ "$dir" = "." ]; then
-            skip=false
-            for s in $STANDALONE_TOP; do
-                [ "$base" = "$s" ] && skip=true && break
-            done
-            if $skip; then
-                # Special case: data.js -> data_base.js
-                if [ "$base" = "data.js" ]; then
-                    link="$SITE/data_base.js"
-                    if [ ! -L "$link" ]; then
-                        errors+=("missing symlink: site/viewer/lib/data_base.js -> $src_file")
-                    fi
-                fi
-                continue
-            fi
-        fi
-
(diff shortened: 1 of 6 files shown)


brayniac (approved):
Reviewed the full diff and independently reproduced the fix. The `charts/boxplot.js` symlink is byte-identical to a fix I verified end-to-end in headless Chrome (before: splash stuck at "Loading…", 404 on `lib/charts/boxplot.js`, zero canvases; after: 155 chart cards render in ~1s). The resolution-based (`-e`) guard correctly handles directory-symlink coverage and would have caught this outage, and the skill/doc corrections match what I independently found (data.js is now a plain symlink; no data_base.js). All checks green. LGTM.

PR #1022, opened 2026-07-22
docs: mean-vs-median decimation discussion (journal) + jitter/timestamp notes

thinkingfish:
## What

Docs-only. Three pieces:

**1. New journal entry — [mean vs. median for the decimated line](docs/journal/2026-07-21-decimation-mean-vs-median.md)** (OPEN, discussion). Records the tradeoff so it isn't re-derived:
- The shipped rationale for the **median** line: robust typical level (a `[80…400]` bucket has median 102.5, mean 134.5 — above the inner band's top edge), the `p25 ≤ median ≤ p75` line-inside-band invariant, and the exact min–max envelope carrying the spikes ("lossless for extremes").
- The **mean**'s counter-argument: conservation — `mean × bucket-width = Σ samples`, so area-under-line reads as true total work; the median under-counts bursty workloads exactly where that reading is natural (rates/counters).
- An explicit de-conflation: the chart-header **Mean/Total toggle is cross-series aggregation**, not the per-bucket time statistic.
- Open questions (per-bucket mean in the display wire, tooltip "(median)" qualifier, whether notebook/compare stats inherit median semantics — flagged as unverified) and decision criteria. Indexed in the journal README, mirrored into `docs/backlog.md` per journal conventions.

**2. `parquet_metadata.md` — raw vs. grid-snapped timestamps.** Why sampling jitter is only observable through `MetricsSource::sample_timestamps()` (the TSDB snaps timestamps on ingest, so PromQL-path data is structurally flat), and why `parquet combine` **permanently destroys** the jitter signal (on-disk quantization) — jitter analysis needs the original single-source file.

**3. Backlog — jitter distribution side panels.** CDF of inter-sample interval + PDF of jitter deviation, rendered beside the timestamp jitter chart (below on narrow screens), summarizing the current zoom selection.

## Related

- #1021 (band swatch row) shipped the labeling half of the median-line story — the swatch names the line "median".
- #1017 uncertainty bands prompted the discussion.


brayniac:
As a note, I think we need to support both raw and snapped timestamps. The core to this effort is trying to display data more honestly while trying to give good hints at behavior. As we take on more cases where samplers are decoupled or we are ingesting external streams, the artifice of aligned phases for observations becomes more problematic. Differing sample frequencies creates additional issues here. Further discussion to be had here, but my feeling is that moving away from snapped timelines is likely the better path for simple queries.

docs/parquet_metadata.md
```diff
@@ -107,6 +107,18 @@ bogus default and the baseline falls back to the average of the real intervals.
...
 
+**Raw vs. grid-snapped timestamps.** Jitter is only observable on the raw
+`timestamp` column. The viewer's query pipeline (TSDB/PromQL) snaps sample
+timestamps to the `sampling_interval_ms` grid on ingest, so anything read
+through a query — including a hypothetical synthesized inter-sample-delta
+metric — is structurally flat and cannot express jitter. The jitter chart
+therefore bypasses the query path entirely and reads un-snapped timestamps via
+`MetricsSource::sample_timestamps()` (served as `/api/v1/timestamps`). Any
+future feature needing sub-interval timing fidelity must do the same. Note
+also that `parquet combine` quantizes timestamps to the common grid **on
```
brayniac: Parquet combine should not be making wider tables anymore? We've introduced the new .rez format with split tables. It's intended that things are split across multiple tables per Rezolus. The combine should no longer do anything with timestamps? (This might not be addressed yet)

brayniac (approved):

thinkingfish:
Looks like an older local doc got picked up. Let me reconcile

– Yao ***@***.***)

PR #1033, opened 2026-08-11
docs: streaming segmented .rez writer design (journal)

brayniac:
## Summary

- **Journal entry (2026-08-11):** design for bounded-time `.rez` finalization, landed pre-build per the journal convention. Driver: a client stops the recorder on docker tear-down (~10s SIGTERM→SIGKILL grace), but finalize cost currently scales with recording length. Scope: rezolus + metriken (metriken-query segment-aware source).
- **Design:** a dedicated writer thread streams sealed per-sampler parquet segments (`<dir>/<sampler>/<seq>.parquet`; byte-first thresholds via incremental in-memory accounting) directly into the output tar during recording, with a two-sync `manifest.json` checkpoint after every seal batch — an unclean kill or power loss leaves a `.partial` readable up to the last synced checkpoint, distinguishable via a per-recording `complete` marker. Row timestamps become **monotonic by construction** (wall anchor at start + monotonic elapsed; anchor recorded in the manifest, both clocks captured per tick, per-row `:wall_offset` observations + per-checkpoint summary series make drift/steps row-locatable metadata) — a recorder-side NTP step can no longer bake non-monotonic time into immutable segments. Finalize = seal tails + final manifest + footer + rename — bounded by thresholds, not length. Manifest v2 (`files: Vec<String>`, v1 archives still readable; forward version gate). **Reading is segment-aware in metriken-query (in scope): footer-only open per segment, lazy row-group decode, one spliced timeline per series with a column-identity conflict policy — merge-at-read rejected; no read-path regression is a GO criterion.** Precise truncation tolerance in `read_archive_reader`. Also removes the `.rez`-mode msgpack spool double-write.
- **Backlog sync:** compactor (space + v1-compat downgrade path; optimization, not a read-speed requirement), metric-identity column keys, seal-threshold constants, classic-parquet fast finalize (by-design).

## Adversarial review (pre-build, on the design)

Three parallel adversarial subagents attacked the design against the code and vendored crates (tar-0.4.46, histogram-1.5.0, metriken-query). All findings verified and folded into the entry; none discarded.

**Fixed in the design:** two wrong claims corrected (merge cost "same as reading today" — today's open is footer-only lazy, resolved by pulling a segment-aware metriken-query source into scope and rejecting merge-at-read; "disjoint increasing timestamps" — `last_key` orders keys, not timestamps: follow-up made row stamps monotonic by construction); single-sync checkpoint had a power-loss ordering hole → two-sync protocol + parent-dir fsync; tar-rs silently yields short entries on mid-data truncation → length-checked tolerance rule; no manifest before the first seal batch → initial empty manifest as first tar entry; `File::create` truncated pre-existing outputs at t=0 → `.partial` + rename; top-level `complete` couldn't survive `combine` → per-recording; tool output on segmented input was broken → `write_archive_bytes` canonicalizes the index (tools pass segments through byte-identical); numeric-id columns remap on agent restart → split-on-conflict identity policy (+ latent `push_row` values/windows desync flagged); "estimated segment bytes" unimplementable → incremental accounting in `push_row`; panic hook exits before unwind → panic-free writer contract; writer joined on every exit path (no-data returns leaked a stub); age seals tick-driven + scrape/probe timeouts (hung endpoint stalled seals and SIGTERM).

**Verified safe:** duplicate tar names last-wins (our reader, GNU/bsdtar, Python tarfile); hand-off types all owned/`Send`; `filter --samplers` matches sampler names not paths; read path entry-order-agnostic; in-progress reads sequential (no mmap).

**Accepted risks (recorded):** disk-bottleneck backpressure can still eat the grace window (bound formula in GO criteria, queue depth 1–2); mid-archive corruption presents as truncation (under-reports, never fabricates); v1 binaries can't see `complete` on compacted recovered archives; agent-side clock step-back still regresses dedup keys (gap, not corrupt math — pre-existing).

## Test plan

- [x] Docs-only change; entry's code references verified against current `main`
- [ ] Implementation follows in a separate PR with the entry's testing plan (truncation geometries, segmented-source conflict policy, read-path parity — footer-only open asserted, monotonic stamping, canonicalization, initial-manifest sniff, no-data unlink, round-trip)


brayniac:
Superseded by #1041, which carried this journal entry in along with the implementation.

The design landed here first as intent-before-build, then #1041 branched from it and evolved it in place — folding in the adversarial-review amendments, the implementation outcome, and the Linux fleet measurements. All of that is now on `main` at `5de241d9`.

Merging this branch today would be a regression: `git diff main...journal-rez-streaming-writer` is +4/−198, i.e. it would delete the outcome and measurement sections that main already carries.

Closing rather than merging. The design record lives at `docs/journal/2026-08-11-rez-streaming-writer.md` on main, and the follow-on effort is #1042.

PR #1064, opened 2026-08-17
perf(rez): stop writing sidecar columns for metrics with no window

brayniac:
## Summary

Every metric in a `.rez` table got a `:window_begin` and `:window_width` column whether or not it had an acquisition window — and most don't. Whole samplers (`cpu_perf`, `cpu_bandwidth`, `cpu_frequency`, `cpu_l3`, `cpu_dtlb`, `cpu_branch`) are windowless, as are the per-cgroup and per-task halves of `cpu_usage`. Over half of every wide table was carrying nothing.

The writer now emits the pair only for columns that actually have a window. Implements the first of the three changes scoped in [the design entry](https://github.com/iopsystems/rezolus/blob/main/docs/journal/2026-08-17-window-sidecar-cost.md) (#1063); the two agent-side ones are untouched and still gated on profiling the counter sweep.

## Measured against a live 32-core agent

| table | before | after | |
|---|---|---|---|
| `cpu_usage` | 6,436 | **2,963** | 2.17× |
| `syscall_counts` | 1,687 | **603** | 2.80× |
| `scheduler_runqueue` | 1,198 | 606 | 1.98× |
| `cpu_perf` | 631 | **219** | 2.88× |
| **all tables** | **12,893** | **6,156** | **2.09×** |
| 60 s archive | 36.9 MB | **29.3 MB** | −20.6% |

Same agent, same settings, before/after binaries built from `c07767fb` and this branch.

## The behaviour change, which is the real reason to do this

Absence and all-nulls read identically — both readers resolve a window per row through an `Option` that a missing column leaves `None`, the path every pre-windows parquet already takes. With one deliberate exception.

An all-null sidecar **still resolves a window**, landing on `(base, base)`: zero width, asserting the observation happened at an exact instant we do not know. Run through the rate bounds that doesn't collapse to the nominal — it fabricates an asymmetric band. On this branch's own before-binary:

```
sum(rate(cpu_cycles[1m]))
  First: 3269950434  [3269950434.000000, 8190749101.498776]
```

`cpu_cycles` carries no window at all, and the recorder was reporting a 2.5× uncertainty spread invented from nothing. After, the same query returns its 59 points with no band — which is what "we do not know when this was acquired" should look like.

Worth knowing before merge: the six windowless samplers lose their brackets in `mcp query` output and their bands in the viewer. That's the intended correction, not a regression.

## Test plan

- [x] `cargo test` — 590 + 7 + 5 green, including the end-to-end hindsight dump suite driving the real daemon
- [x] `cargo clippy --all-targets` clean, zero build warnings, `cargo fmt --check`
- [x] New: a windowless column emits no sidecars; a column with **any** windowed row keeps them; windowedness differing between segments of one table; a column-count assertion pinning the saving
- [x] End-to-end on a live agent — column counts, archive size, and `mcp query` band behaviour for a windowed and a windowless metric
- [x] Verified both readers treat absence as all-null: `metriken-query` `parse_schema` attaches sidecar indices as `Option` (`parquet.rs:1416`), and in-tree `read_table_parquet` uses `unwrap_or_default()`

## Note on one test that moved

Three dump tests failed after this change, and the fixture check is the one that fired: it requires each dump to outlast a tick period, since a dump shorter than `INTERVAL` couldn't distinguish a writer that keeps running from one that pauses for every dump. Dump cost scales with archive size, so narrowing the tables took the average dump to 91 ms against a 100 ms tick.

`WIDE` is recalibrated with a comment explaining the coupling, so whoever next makes tables narrower finds out why that constant exists rather than rediscovering it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2946,7 +2946,7 @@ dependencies = [
 
 [[package]]
 name = "rezolus"
-version = "5.17.1-alpha.15"
+version = "5.17.1-alpha.16"
 dependencies = [
  "allan",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ wasm-bindgen = "0.2.120"
 
 [package]
 name = "rezolus"
-version = "5.17.1-alpha.15"
+version = "5.17.1-alpha.16"
 description = "High resolution systems performance telemetry agent"
 edition = "2021"
 license.workspace = true
--- a/src/hindsight/buffer.rs
+++ b/src/hindsight/buffer.rs
@@ -848,6 +848,11 @@ mod tests {
             buf.ingest(&s, ts, 0).unwrap();
             buf.maintain().unwrap();
         }
+        // Everything below reads the file through its OWN connection, so it
+        // sees what the writer has committed rather than what this buffer has
+        // handed off. Without the barrier the last ticks may still be in
+        // flight and the row count comes up short.
+        buf.sync().unwrap();
 
         assert_eq!(detect_rez_format(&path).unwrap(), RezFormat::V3Sqlite);
 
--- a/src/recorder/rez.rs
+++ b/src/recorder/rez.rs
@@ -259,19 +259,37 @@ fn table_to_batch(table: &RezTable) -> Result<(Arc<Schema>, RecordBatch), RezErr
             }
         }
 
-        let (begin, width) = window_offset_columns(&col.windows, &table.timestamps);
-        fields.push(Field::new(
-            format!("{}:window_begin", col.name),
-            DataType::Int64,
-            true,
-        ));
-        arrays.push(Arc::new(Int64Array::from(begin)));
-        fields.push(Field::new(
-            format!("{}:window_width", col.name),
-            DataType::UInt64,
-            true,
-        ));
-        arrays.push(Arc::new(UInt64Array::from(width)));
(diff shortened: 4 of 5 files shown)


brayniac:
Closing: the acquisition-groups arc removed the thing this optimizes.

This PR's premise is that whole samplers are windowless — it names
`cpu_perf`, `cpu_bandwidth`, `cpu_frequency`, `cpu_l3`, `cpu_dtlb` and
`cpu_branch`. All six now declare acquisition groups (29 sampler files do, as
of #1070), so they all carry windows.

More fundamentally, there are no per-metric sidecar columns left to suppress on
the default path. A V3 group table carries a single table-level
`:window_begin`/`:window_width` pair for every metric in it, because a group is
by definition one read with one window. The column-count problem this PR
measured — "over half of every wide table was carrying nothing" — cannot arise
in that layout.

The measured win is superseded too. This PR got a 2.09× column reduction and
−20.6% archive size; V3 measured **0.386×** archive size against the same V2
baseline at a 50 ms interval, along with 0.52–0.60× query latency. Same
direction, considerably further.

What remains is `snapshot_format = "v2"`, now an escape hatch for consumers
that cannot read V3 (#1076 made V3 the default). Carrying a V2-only writer
optimization for that path is not worth the maintenance, and the sidecar-cost
design entry it implements (#1063) is now history rather than a plan.

The branch stays if anyone wants the measurement methodology — the
before/after column-count harness was genuinely useful and informed the
split-table gate that followed.

PR #1065, opened 2026-08-18
feat(parquet): split-table read-cost measurement gate

brayniac:
## Summary
- Adds dev scaffolding to measure the read cost of the split-table (per-acquisition-group) `.rez` layout before committing to any production writer changes — the go/no-go gate for the acquisition-groups design direction recorded in `docs/journal/2026-08-17-window-sidecar-cost.md` (see the new addendum there).
- New hidden subcommand `rezolus parquet split-groups -i wide.rez -o out.rez [--wide]`: rewrites a finalized v3 recording into per-acquisition-group tables (`<sampler>/acq_<member>`, `<sampler>/f_<metric>`), inferring cohorts from agreeing `:window_begin` values (null-pattern tolerant) and grouping windowless metrics by base-metric family. `--wide` produces a provenance-matched identity re-encode so the comparison isolates layout, not encoding. Refuses inputs with live WAL rows rather than silently truncating them.
- `scripts/split-rez-measure.sh`: three same-filesystem arms (orig / wide-rt / split), all-of metric preflight, untimed warm-up, round-robin timed reps of `parquet metadata` and single-group `mcp query`, TSV output; temp dir preserved on failure.
- Only production-path change is one visibility keyword (`segment_writer_props` → `pub(crate)`). The splitter module is explicitly throwaway — deleted when the real grouped writer lands.

## Test plan
- [x] 13 unit tests in `split_groups.rs` (cohort partitioning, null-pattern merging, histogram sidecar cohorting, wide-mode byte-level identity, container round-trip, live-WAL refusal, incomplete-recording preservation)
- [x] Full `cargo test` suite green; clippy/fmt clean for touched files
- [x] End-to-end exercised on generated multi-segment v3 fixtures: split output opens via `RezReader`; `describe-metrics` catalog identical between wide re-encode and split arms (no phantom window metrics); group names stable across segments when a cohort goes quiet
- [x] Measurement run on a Linux host (300 s / 50 ms recording, steady workload) + results journal entry: split is 0.55–0.68× on query latency, 0.63× on bytes vs the wide re-encode — see `docs/journal/2026-08-18-split-table-read-cost-gate.md`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2946,7 +2946,7 @@ dependencies = [
 
 [[package]]
 name = "rezolus"
-version = "5.17.1-alpha.15"
+version = "5.17.1-alpha.16"
 dependencies = [
  "allan",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ wasm-bindgen = "0.2.120"
 
 [package]
 name = "rezolus"
-version = "5.17.1-alpha.15"
+version = "5.17.1-alpha.16"
 description = "High resolution systems performance telemetry agent"
 edition = "2021"
 license.workspace = true
--- a/docs/journal/2026-08-17-window-sidecar-cost.md
+++ b/docs/journal/2026-08-17-window-sidecar-cost.md
@@ -179,6 +179,10 @@ work being self-evident from the loop bound.
 
 ## Addendum (2026-08-17): split tables reopened, gated on measurement
 
+> **Update 2026-08-18:** the gate ran — see
+> [split-table read-cost gate](2026-08-18-split-table-read-cost-gate.md).
+> Queries 0.55–0.68×, archive 0.63× versus the wide layout.
+
 The "Split tables per cohort" rejection above is superseded by the
 acquisition-groups design (spec in the local, untracked
 `docs/superpowers/specs/2026-08-17-acquisition-groups-design.md`): the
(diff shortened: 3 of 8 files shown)


brayniac:
Measurement gate results (docs/journal/2026-08-18-split-table-read-cost-gate.md):

Input: 300 s / 50 ms recording from a live 30-sampler agent on a 32-core x86_64 host under a steady workload — 314.6 MB, 309 segments, 26 tables. Three arms in one tmpfs dir, warm-up pass, 7 round-robined reps.

| | wide-rt | split | ratio |
|---|---|---|---|
| archive bytes | 314.6 MB | **197.7 MB** | **0.63×** |
| segments / tables | 309 / 26 | 1,723 / 102 | 5.6× / 3.9× |
| `parquet metadata` | 4 ms | 9 ms | 2.25× |
| per-CPU sum-rate query | 3,332 ms | **1,816 ms** | **0.55×** |
| cgroup by-name rate query | 2,616 ms | **1,780 ms** | **0.68×** |
| scheduler sum-rate query | 2,625 ms | **1,773 ms** | **0.68×** |

Query answers verified identical across arms (sorted output byte-identical; split arm lacks bounds annotations per the stated reader caveat). The segment-count fear inverted: per-segment open cost tracks schema width, and narrower tables more than pay for 5.6× the segments.

**Recommendation: gate passes** — split layout is both smaller and faster to query on the same data, with the windowless-family over-splitting handicap against it. Awaiting go/no-go before Stage 2 planning.

PR #1122, opened 2026-09-01
feat(viewer): open a .rez archive in the browser

brayniac:
Follows #1121. The static-site viewer was parquet-only, so it silently failed on every streamed recording — and `.rez` is what `record` writes by default. Verified end to end by running the **shipped WASM bundle** in node against a real two-recording archive:

```
baseline attached: true    experiment attached: true
baseline   filename=redis-web-01   metadata={"source":"redis"}
experiment filename=valkey-web-01  metadata={"source":"valkey"}
```

**Four commits, meant to be read in order** — each stands alone:

1. **`refactor(rez): make the reader half compile without metriken`** — the blocker. `metriken-core` declares a `linkme` distributed slice, and `linkme` gates on a fixed `target_os` list that `unknown` is not in, so anything reaching `metriken` fails to compile for wasm32 outright. A `write` feature (default on) now carries the ingest half; without it the crate is a reader.
2. **`feat(rez): open an archive from bytes`** — a browser has an upload, not a path.
3. **`feat(viewer): open a .rez archive in the browser`** — the payoff.
4. **`fix(viewer): say when a browser-opened archive reads short`** — see "the one honest gap" below.

## Three types moved into the archive (commit 1)

Each was already owned by the wrong crate; the wasm constraint just made it visible.

- **`rez::window::Window`** — the acquisition window a segment stores. A file format should not need the agent's metrics registry to hold two timestamps.
- **`rez::schema::{GroupSchema, MetricDesc}`** — a group's membership as carried in a WAL row. **rmp-serde writes a struct as a positional array, so the field order IS the on-disk encoding**: a reordered field would not fail to compile, it would write a WAL a released reader decodes into the wrong fields. Pinned byte-for-byte against the producer's type, plus the FNV-1a-128 schema hash the two sides compare across the WAL.
- **`rez::rez::{Cell, CellValue}`** — the builder's input. Rows reach a table from an agent snapshot *and* from the archive's own WAL, and only the first has snapshot values to borrow. The builder spoke `Entry`, so the WAL path rebuilt `Counter`/`Gauge`/`Histogram` values it had *just decoded*, purely to have something to borrow — that is what put a metrics-registry dependency on the read path. Histograms are still rebuilt through `histogram::Histogram::from_buckets`, so a malformed WAL cell is still caught there.

## Things that break in a browser without failing to compile

- **`std::thread::spawn`** compiles for wasm32 and panics at runtime. The reader's parallel footer probe would have built clean and aborted on the first archive opened; it has a sequential wasm arm now.
- **`sqlite3_deserialize` refuses a `.rez` as-is.** The file header says WAL mode, an in-memory database cannot do WAL, and SQLite answers `unable to open database file`. Rewriting header bytes 18/19 to "rollback journal" — what `PRAGMA journal_mode = DELETE` would have persisted — fixes it.
- **A raw copy of an archive a writer still holds** is a valid SQLite database with none of the archive in it. That surfaced as `no such table: recordings`, which reads like corruption; it now says what happened.

## The one honest gap (commit 4)

A `.rez` carries its **own** unsealed rows in a `wal` TABLE inside the file, and those travel with an upload — `materialize_wal_tail` runs in wasm, and a test pins that a `hindsight`-style snapshot's live tail survives the trip through bytes.

SQLite's **`-wal` sidecar** is the part that cannot: a separate file holding pages committed but not yet checkpointed into the archive. `rezolus view archive.rez` opens by path with the sidecar beside it and reads them, so the same recording can read short in the browser and complete from the CLI, by up to a checkpoint's worth of ticks. It only differs for a copy taken from under a running recorder — a clean exit checkpoints, a `hindsight` snapshot is a consistent single file — but rendering short in silence looks exactly like a recording that stopped early. The archive already knows (`complete` stays 0 until finalize) and `RezReader` only warned through `tracing`, which in wasm goes nowhere; `RezReader::complete()` carries it out and the registry reports it.

## Browser behaviour

A 2-recording archive fills both A/B slots — the same mapping `rezolus view` makes. Loading one arm into one slot would show an A/B archive as a single capture with nothing on screen saying half of it is missing. Above two, the first two are shown and the rest are **named** in `WasmCaptureRegistry::notices()`, which the frontend logs.

Aliases come from `dashboard::capture_alias::discriminating_alias_key`, moved out of `src/viewer/` so both viewers name captures identically — one viewer showing `web-01` where the other shows `baseline` is the kind of divergence nobody reports.

Save as Report stays parquet-only (it re-encodes a projection of the capture's original file, and an archive is many tables); `Viewer::can_save` reports that rather than handing back an empty download.

## Costs and testing

Bundle **4.5 → 6.1 MB raw, 1.4 → 2.0 MB gzipped** (SQLite), measured both ways.

`crates/viewer` becomes `["cdylib", "rlib"]` so its behaviour is testable at all — wasm-bindgen's generated code compiles natively, and none of the JS interop runs in these tests. That adds a `cargo test -p viewer` CI step, since the workspace test run excludes it. Also new: a CI check that the reader still builds for wasm32 (nothing else catches a metriken type creeping back onto the read path), and `tests/wasm_viewer_rez_archive.test.mjs`, which drives the built bundle from node.

172 tests in `rez`, 9 in `viewer`, 6 alias tests moved to `dashboard`, 249 JS tests green, clippy clean across the workspace.

**macOS note:** `crates/viewer/build.sh` has never worked with Apple's `clang`, which has no wasm32 target (zstd, and now SQLite, are compiled from C). Homebrew LLVM does — `CC_wasm32_unknown_unknown=/opt/homebrew/opt/llvm/bin/clang`. Documented in CLAUDE.md.


brayniac:
Added a fifth commit — `feat(recording): snapshot a .rez that is still being written` — after measuring the gap I described in "the one honest gap" above. I had been hand-waving it as "up to ~4 MiB"; it is bigger than that.

**Measured**, 2000-tick recording, 40 counters/tick:

| | |
|---|---|
| archive on disk | 1.9 MB |
| SQLite `-wal` sidecar | **4.2 MB** |
| newest row, read in place | tick 2000 |
| newest row, from a plain `cp` | tick 1877 |

**123 ticks behind — ~2 minutes at a 1s interval**, with a sidecar larger than the archive itself. For incident analysis that is precisely the interesting window, so a warning was not a sufficient answer.

`rezolus recording snapshot live.rez -o incident.rez` reads through the sidecar and writes one consistent file, without pausing the writer. It is `VACUUM INTO` under the hood — the same primitive `hindsight`'s dump already used, which was reachable only from hindsight's own trigger and not for an arbitrary live archive. The two messages that previously said "snapshot it with `rezolus hindsight`" now name this instead, since it works whether or not hindsight is involved.

The test asserts the **difference**, not merely that the snapshot opens: the snapshot must reach the last committed tick and the plain copy must not. If some future change made a bare copy complete, that half fails and this command's rationale wants revisiting — which is exactly when someone should look at it.

PR #1125, opened 2026-09-01
feat(rez): expose a recording's tables as composition sources

brayniac:
A `.rez` cannot currently take part in a **cross-artifact** query.

Merging several recordings into one labelled multi-source — so a single query
spans them and the answer can be sliced by which recording it came from — needs
each participant as a `metriken_query::CompositionSource`. That is built only
`From<&ParquetReader>` / `From<&SegmentedParquetReader>`, and a `RezReader` is
neither: it is a composition of per-sampler tables, with no public way to reach
them.

This is the blocker for reading `.rez` in systemslab, whose core query path
merges N job artifacts into one reader with `artifact_id` / `job` labels
injected, so one PromQL query can span jobs
([design](https://github.com/iopsystems/systemslab/blob/main/docs/superpowers/specs/2026-09-01-rez-format-design.md)).

## The change

`crates/rez` already does exactly this privately, for its own unions:

```rust
fn union_child(&self) -> UnionChild {
    match self {
        TableReader::Single(r) => UnionChild::from(r),
        TableReader::Segmented(r) => UnionChild::from(r),
    }
}
```

`TableReader::composition_source()` sits next to it — same match, same borrow —
and `RezReader::composition_sources()` maps it over the recording's tables.

## Why two composers rather than one

They have opposite disjointness rules, and both are needed:

- **`union_child`** feeds `route`'s same-recording union, whose children must
  hold **disjoint** metric names. A collision there is a producer/archive bug
  and `try_new` refuses it.
- **`composition_source`** feeds the cross-artifact builder, where every child
  holds the **same** names and must stay distinct under its own injected label
  — the failure mode being one silently winning.

## Cost, stated in the API

`composition_sources()` **opens every table**, and the doc comment says so.
Routing deliberately builds only the tables a query names (`SamplerReader::reader`
is a lazy `OnceLock`) because a typical query touches a small fraction of a large
archive — the header notes ~1.37 ms per segment over 418 segments, of which a
typical query needs 11%. A composed source cannot work that way: the builder
dispatches across its children by metric name and needs them all up front. So
the guidance is compose once and reuse, not compose per query.

A table whose segments cannot be parsed is skipped rather than poisoning the
composition, matching `SamplerReader::reader`.

## Verification

- `cargo test -p rez --features test-support` — **174 passed**, 0 failed (173
  pre-existing plus the new one).
- `cargo clippy -p rez --all-targets --features test-support` — clean.
- `cargo +nightly fmt --all -- --check` — clean.

New test `composition_sources_merge_recordings_under_distinct_labels` writes two
v3 archives, composes both into one `ParquetBuilder` under `job=a` / `job=b`, and
asserts both arms survive distinguishable — neither merged nor dropped.

Note `cargo clippy -p rez --all-targets` *without* `--features test-support`
fails on the `write_rez_fixture` example, which references
`rez::recorder_tests_support`. That reproduces on a pristine `main` and is
untouched here.

https://claude.ai/code/session_012Jtzg9VbUHX56UpQ8JEJ83

### DIFF
--- a/crates/rez/src/reader.rs
+++ b/crates/rez/src/reader.rs
@@ -427,12 +427,43 @@ impl RezReader {
     ///
     /// A table whose segments cannot be parsed is skipped rather than
     /// poisoning the composition, matching `SamplerReader::reader`.
-    pub fn composition_sources(&self) -> Vec<CompositionSource> {
-        self.tables
+    ///
+    /// # Errors
+    ///
+    /// Refuses a reader that **flattens several recordings**, which is what
+    /// [`open_with_pool`](Self::open_with_pool) produces. Every recording of a
+    /// multi-recording archive holds the same sampler names, so composing a
+    /// flattened reader hands the builder several children carrying the same
+    /// metric names under one label — and one silently wins, losing an arm's
+    /// data entirely. That is the exact failure this composer exists to
+    /// prevent, so it refuses rather than warns, mirroring `route`, which
+    /// groups owners by `(recording, sampler)` for the same reason.
+    ///
+    /// Open with [`open_recordings`](Self::open_recordings) instead: one
+    /// reader per recording, each composable under its own label.
+    pub fn composition_sources(
+        &self,
+    ) -> Result<Vec<CompositionSource>, Box<dyn std::error::Error>> {
+        let mut recordings: Vec<usize> = self.tables.iter().map(|t| t.recording).collect();
+        recordings.sort_unstable();
+        recordings.dedup();
+        if recordings.len() > 1 {
+            return Err(format!(
+                "cannot compose a reader flattening {} recordings: they share sampler \
+                 names, so composing them would let one recording's series silently \
+                 replace another's. Open the archive with `open_recordings` and compose \
+                 each recording under its own label.",
+                recordings.len()
+            )
+            .into());
+        }
+
+        Ok(self
+            .tables
             .iter()
             .filter_map(|t| t.reader())
             .map(TableReader::composition_source)
-            .collect()
+            .collect())
     }
 
     /// Whether this recording holds no tables at all.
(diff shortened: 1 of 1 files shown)


brayniac:
Both findings were right, and the first was a real bug. Fixed in 67f21cb0.

**1. Silent merge on a flattened reader.** Confirmed exactly as described.
`open_with_pool` flattens every recording's tables into one vec and
`composition_sources` mapped over it unconditionally, so a 2-recording archive
handed the builder two children with the same metric names under one label and
one silently replaced the other.

What makes this clearly wrong rather than debatable: `route` already refuses the
same hazard, grouping owners by `(recording, sampler)` rather than sampler
alone, pinned by
`multi_recording_same_sampler_query_errors_instead_of_silently_dropping_one_recording`.
The composition path bypassed that protection while its own doc claimed to
provide it.

`composition_sources` now returns `Result` and refuses when its tables span more
than one recording, naming `open_recordings` as the entry point that works —
refusing rather than documenting, per the suggestion and the surrounding code's
habit.

`composition_sources_refuse_a_flattened_multi_recording_reader` pins it, and
also asserts the same archive opened per recording still composes, so the
refusal is about flattening rather than the fixture. I verified it fails with
the guard stubbed out — it reproduces the bug rather than merely passing.

The happy-path test now opens with `open_recordings` instead of
`open_with_pool`, so it models the entry point composition actually requires.
That was the thing steering a reader toward the broken usage.

**2. Displaced doc comment.** Confirmed and restored. The new test landed
between `a_table_evicted_between_probe_and_query_does_not_panic` and its
rationale, so that three-paragraph explanation of the `.expect(...)` panic and
why hindsight's retention makes it reachable ended up on an unrelated test,
running into its text mid-paragraph, and left the eviction regression with
nothing.

`cargo test -p rez --features test-support`: 175 passed, 0 failed. clippy and
`cargo +nightly fmt --all -- --check` clean.

One API note: `CompositionSource` is opaque and not `Debug`, so the test matches
on the `Result` rather than using `expect_err`. Deriving `Debug` upstream would
be a small ergonomic win if this pattern recurs.

brayniac:
## Adversarial review

Ran two adversarial subagents over `upstream/main...HEAD` — one on the guard's
soundness across every construction path, one on composition semantics against
`MultiParquetSource`/`UnionMetricsSource`. Findings and disposition:

### Fixed (120d064a)

**The guard was sound only by coincidence.** It counted distinct
`SamplerReader::recording` values, but that field's meaning is not uniform:
`from_v3_db` numbers recordings archive-globally, `from_recordings` per call. It
worked only because `open_with_pool`'s tar branch passes every recording in one
call. Rebuilding that branch on `open_recordings` — the obvious cleanup, since
the two branches are near-duplicates — would give every table `recording == 0`
and silently defeat the guard *with its test still green*. `RezReader` now
carries a `recordings` count set at construction; it is a required field, so a
future constructor cannot omit it.

**The v3 branch was untested.** The refusal test built a v2 tar archive, so it
covered only `from_recordings`. v3 reaches its count by a different route and is
the branch that matters — `parquet combine` emits v3, and `record --endpoint a
--endpoint b` writes multi-recording v3 directly. Added the twin; both refusals
verified to fail with the guard stubbed out.

**Empty compositions returned `Ok(vec![])`.** Skipping every unopenable table
left an evicted arm indistinguishable from a flat one — the same silent per-arm
loss the refusal exists to prevent. Now an error.

**The stated failure mode was wrong**, in the last commit message and the doc.
I said one recording's series "silently replace" another's. That is
`UnionSource`'s first-wins — a different composer. `ParquetBuilder`
*concatenates* ("same (metric, label) pairs in multiple files produce duplicate
series"), so the real outcome is duplicate indistinguishable series and a
silently doubled aggregate, with histograms interleaving into one series index.
The refusal stands; its reasoning now matches the code, so a maintainer who
checks it does not find it false and weaken the guard.

### Documented, not fixed — both are real and neither belongs here

**`open_recordings` labels are not unique.** Two endpoints inferring the same
`source`/`host` produce identical label sets and the recorder only warns. A
caller labelling arms from recording labels reproduces the merge one level up,
where this method cannot see it. The doc now states the caller owns label
uniqueness.

**Composition tolerates name-sharing `route` refuses.** `route` rejects a query
whose metrics span tables sharing a name (`try_new`). This composition cannot:
it is eager and whole-archive, so refusing would reject an entire archive over a
name that may never be queried — and `gpu_nvidia`/`gpu_amd_smi` genuinely both
publish `gpu_utilization`. Consequence documented, including that the series
stay separable by their `sampler` label.

### Verified safe

**Acquisition windows survive composition.** Windows are per-series fields
resolved inside each child's own read path; `MultiParquetSource` only injects
labels and never touches them, and `SegmentedParquetReader` merges windows
within its own segments. A `rate()` band is as precise composed as through
`route`.

### Known limitation, filed against the consumer rather than this PR

`RezReader::query_range_opts` injects `cross_cadence_eval_timestamps` so a query
spanning a fast and a slow sampler evaluates at the slow table's real rows. That
policy lives on `RezReader`, not on any `DataSource`, so a *composed* reader
loses it — on the branch's own `cross_cadence_rez()` fixture, 2 correct points
become 9, seven at instants the slow sampler never read.

This is inherent to composing, not a defect introduced here, and it does not
affect the common consumer query (one sampler across several artifacts, uniform
cadence). systemslab will avoid it by querying a single artifact through
`RezReader` directly and reserving composition for genuinely multi-artifact
views. If composed cross-sampler queries later need full fidelity, the fix is to
expose the eval timestamps so the caller can set them on the composed reader's
`QueryOptions` — worth a follow-up issue rather than growing this PR.

### Gate

`cargo test -p rez --features test-support`: **176 passed**, 0 failed. clippy
and `cargo +nightly fmt --all -- --check` clean.

PR #1173, opened 2026-09-07
feat: Intel GPU sampler, and per-GPU charts that tell two vendors apart

yangxi:
Adds an Intel GPU sampler and brings the viewer's per-GPU charts in line with
what multi-vendor hosts actually look like.

## The sampler

`gpu_intel_pmu` reads the i915/xe PMU via `perf_event_open`, covering both
integrated GPUs and discrete Arc cards. Each GPU registers its own PMU, which is
discovered from sysfs along with the engine set that GPU exposes — the perf type
is assigned at boot (23, 24 and 32 observed across two hosts) and the engine set
differs per part, so nothing is hardcoded.

| metric | type | source |
|---|---|---|
| `gpu_engine_busy_time` | counter, ns | perf event `<engine>-busy` |
| `gpu_frequency_sample{frequency=actual\|requested}` | counter, MHz | perf events |
| `gpu_memory{state=used\|free}` | gauge, bytes | `DRM_IOCTL_I915_QUERY` |

All events for a GPU open as one perf group, so a single read returns every
counter with no skew between engines. Reads take a driver lock rather than being
an mmap load, so the sampler runs on its own bounded cadence (default 1s) via
`spawn_blocking` — a documented departure from principle 10, per principle 17.

VRAM comes from the DRM query ioctl, since the PMU has no memory counter. It is
discrete-only: an integrated GPU has no device-local memory region and publishes
nothing rather than passing host RAM off as VRAM. Without `CAP_PERFMON` the
kernel reports all memory as unallocated instead of failing the ioctl, so the
sampler suppresses the series rather than publishing a plausible "0 used".

## Viewer

Per-GPU charts grouped `sum by (id)`, which collapses two GPUs that share an id
into one summed line. Every vendor's sampler numbers its devices from 0, so a
host with an NVIDIA card and an Intel iGPU has two GPUs both labelled `id="0"`.
17 queries grouped that way; they now group by `(id, vendor)`.

Three related fixes fell out:

- **Heatmap rows were indexed by `parseInt(id)`**, so two GPUs at `id=0` landed
  on the same row, one overwriting the other. That is data loss, not just a
  labelling problem. Rows fall back to result order when ids are not unique and
  dense, and each carries its own label.
- **`resolveStyle` required ≥2 series for a heatmap**, making chart *type* depend
  on how many devices happened to be present — a "Per-GPU" chart drew a heatmap
  on a two-GPU host and a line chart on a one-GPU host.
- **Seven vendor-exclusive metric names now pin their vendor.** The eight shared
  names are deliberately left unfiltered; pinning those would hide other vendors'
  GPUs.

`gpu_apple` gains `vendor = "apple"` — it was the only sampler publishing none,
leaving the `(id, vendor)` invariant asymmetric across platforms.

## systeminfo

`GpuSummary` dropped five of `hwinfo::Gpu`'s eleven fields at the summary
boundary, including `pci_bus_id`. All five now flow through. The NVIDIA `/proc`
fallback moves into the nvidia backend, which fixes a real bug: it fired on the
*whole* GPU list being empty, so an AMD+NVIDIA host with no NVML skipped the scan
entirely and the NVIDIA GPU went unreported.

## Verification

Hardware-tested on three hosts with both GPUs under load:

| host | GPUs | result |
|---|---|---|
| sky | Arc A770 + UHD 630 | `ccs0` 99.7% @ 2.4 GHz, VRAM tracks allocations |
| hyb | RTX 5080 + UHD 730 | `rcs0` 100%, both vendors in one recording |
| amd | R9700 + Raphael | `gpu_amd_smi` 280 W peak, per-GPU rows separate |

Cross-checked against `intel_gpu_top` reading the same perf events. 747 tests
pass on Linux, clippy clean, viewer symlink guard passes.

## Scope

**i915 only.** The `xe` driver names its PMU the same way but exposes a
different event vocabulary, so it is deliberately not matched — an xe host
reports `Unsupported` with an accurate reason rather than a misleading
permissions error.

## Notes for review

- **The `(id, vendor)` grouping changes every GPU chart**, not just Intel's. It
  is the correct fix, but it is a visible change for AMD and NVIDIA users too.
- **The mixed-vendor conflation has no end-to-end hardware proof.** hyb holds an
  NVIDIA and an Intel GPU both at `id=0`, but no metric *name* is shared between
  them there, so the conflation never fires. Reproducing it needs a discrete
  Intel GPU alongside another vendor; no host available here is that. Covered by
  unit tests.
- **Multi-GT Intel parts are untested** — the sampler would silently drop
  `-gtN`-suffixed globals. Documented in the code, fails safe, no hardware.

## Review round 2

All nine findings from @brayniac's review are addressed (`435100bf`, `66e918fa`),
each reproduced before fixing — including the last bullet of #5, which the
review flagged as unverified and which does reproduce.

The two blocking viewer findings are now verified against **full captures with
every sampler enabled** (1717 columns / 27 samplers, versus 29 for a GPU-only
recording). That matters because both regressions were in *CPU* charts, which an
isolated GPU capture cannot see:

```
             CPU (inferred)                     GPU (explicit override)
sky  (12c)   Busy % (Per-CPU)  n=12  ->heatmap  Compute Busy %  n=1  style=heatmap
hyb  (16c)   Busy % (Per-CPU)  n=16  ->heatmap  GPU %           n=1  style=heatmap
amd  (24c)   Busy % (Per-CPU)  n=24  ->heatmap  GPU %           n=2  style=heatmap
```

`resolveStyle` is untouched again and now has the five tests it was missing;
per-GPU plots opt in via `PlotOpts::style`. Row labels are withheld when they
merely restate their row index, so per-CPU heatmaps keep their "CPU" axis title
and `CPU 2` tooltips.

Correction to an earlier note: the three WASM test failures were **not**
pre-existing on `main`. They were a stale local `site/viewer/pkg/` (gitignored,
dated June) and pass after `./crates/viewer/build.sh` — 273 JS tests, 0 failures.

730 Rust tests on sky, hyb and amd; 277 JS tests; clippy clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3669,6 +3669,7 @@ dependencies = [
 name = "systeminfo"
 version = "0.1.0"
 dependencies = [
+ "libc",
  "libloading",
  "log",
  "serde",
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Rezolus effortlessly tracks the details you need to understand production so you
 - **GPU telemetry** — NVIDIA via NVML and GPM, including per-tensor-pipe
   utilization, plus SM utilization/occupancy, DRAM bandwidth, PCIe throughput, power,
   energy, clocks, and temperature. AMD via ROCm SMI and rocprofiler hardware
-  counters. Intel integrated and Arc GPUs via the i915/xe PMU, with per-engine
+  counters. Intel integrated and Arc GPUs via the i915 PMU, with per-engine
   (render/copy/video/compute) busy time and frequency, plus VRAM usage. Apple GPU
   metrics on macOS.
 - **Container-aware** — per-cgroup CPU cycles/instructions, migrations, syscalls,
--- a/README.md
+++ b/README.md
@@ -49,7 +49,10 @@ Rezolus effortlessly tracks the details you need to understand production so you
   frequency, per core and per cgroup.
 - **GPU telemetry** — NVIDIA via NVML and GPM, including per-tensor-pipe
   utilization, plus SM utilization/occupancy, DRAM bandwidth, PCIe throughput, power,
-  energy, clocks, and temperature. Apple GPU metrics on macOS.
+  energy, clocks, and temperature. AMD via ROCm SMI and rocprofiler hardware
+  counters. Intel integrated and Arc GPUs via the i915/xe PMU, with per-engine
+  (render/copy/video/compute) busy time and frequency, plus VRAM usage. Apple GPU
+  metrics on macOS.
 - **Container-aware** — per-cgroup CPU cycles/instructions, migrations, syscalls,
   and CFS bandwidth/throttling, so you can attribute behavior per container.
 - **Service & inference telemetry** — runtime-loaded templates that turn
--- a/config/agent.toml
+++ b/config/agent.toml
@@ -216,6 +216,29 @@ enabled = false
 # workloads, so use with care on production hosts.
 # gpu_perf_level = "stable_std"
 
+# Produces Intel GPU metrics from the i915/xe PMU via perf_event_open: per-engine
+# busy time and actual/requested frequency, plus VRAM usage via the DRM query
+# ioctl. Covers both integrated GPUs and discrete Arc cards, which register as
+# separate PMUs; each is discovered from sysfs along with its engine set, so no
+# per-model configuration is needed. This is the same data source intel_gpu_top
+# reads. Requires CAP_PERFMON (or kernel.perf_event_paranoid <= 2); disabled
+# automatically on hosts with no Intel GPU or insufficient permissions.
+#
+# VRAM is discrete-only: an integrated GPU has no device-local memory region, so
+# it publishes no gpu_memory series.
+#
+# Note: the PMU values are cumulative counters, so both are consumed as rates —
+# `rate(gpu_engine_busy_time) * 100` is engine utilization in percent, and
+# `rate(gpu_frequency_sample)` is average MHz. Pair them: the PMU reports
+# occupancy, not efficiency, so frequency is what separates "busy but
+# downclocked" from genuinely saturated.
+[samplers.gpu_intel_pmu]
+# Reads touch the driver (a perf group read takes a driver lock; VRAM is an
+# ioctl), so they run on their own cadence rather than at the scrape rate, and
+# are dispatched off the async worker. Default 1s, matching intel_gpu_top's own
+# sampling period. Cached values are served between reads.
+# interval = "1s"
+
 # Memory utilization from /proc/meminfo
 [samplers.memory_meminfo]
 
--- a/config/agent.toml
+++ b/config/agent.toml
@@ -216,7 +216,7 @@ enabled = false
 # workloads, so use with care on production hosts.
 # gpu_perf_level = "stable_std"
 
-# Produces Intel GPU metrics from the i915/xe PMU via perf_event_open: per-engine
+# Produces Intel GPU metrics from the i915 PMU via perf_event_open: per-engine
 # busy time and actual/requested frequency, plus VRAM usage via the DRM query
 # ioctl. Covers both integrated GPUs and discrete Arc cards, which register as
 # separate PMUs; each is discovered from sysfs along with its engine set, so no
--- a/crates/dashboard/src/dashboard/gpu.rs
+++ b/crates/dashboard/src/dashboard/gpu.rs
@@ -114,7 +114,7 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
             PlotOpts::gauge("GPU % (Per-GPU)", "gpu-pct-per-gpu", Unit::Percentage)
                 .percentage_range()
                 .with_row_label("GPU"),
-            "sum by (id) (gpu_utilization) / 100".to_string(),
+            "sum by (id, vendor) (gpu_utilization) / 100".to_string(),
         );
     } else {
         gpu.plot_promql_full(
@@ -139,7 +139,7 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
             )
             .percentage_range()
             .with_row_label("GPU"),
-            "sum by (id) (gpu_memory_utilization) / 100".to_string(),
+            "sum by (id, vendor) (gpu_memory_utilization) / 100".to_string(),
         );
     } else {
         mem_ctrl.plot_promql_full(
@@ -169,7 +169,7 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
             )
             .percentage_range()
             .with_row_label("GPU"),
-            "sum by (id) (gpu_tensor_utilization) / 100".to_string(),
+            "sum by (id, vendor) (gpu_tensor_utilization) / 100".to_string(),
         );
     } else {
         tensor.plot_promql_full(
@@ -194,7 +194,7 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
             )
             .percentage_range()
             .with_row_label("GPU"),
-            "sum by (id) (gpu_sm_utilization) / 100".to_string(),
+            "sum by (id, vendor) (gpu_sm_utilization) / 100".to_string(),
         );
         sm.plot_promql(
             PlotOpts::gauge("GPU SM Occupancy %", "gpu-sm-ocp", Unit::Percentage)
@@ -209,7 +209,7 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
             )
             .percentage_range()
             .with_row_label("GPU"),
-            "sum by (id) (gpu_sm_occupancy) / 100".to_string(),
+            "sum by (id, vendor) (gpu_sm_occupancy) / 100".to_string(),
         );
     } else {
         sm.plot_promql_full(
@@ -249,12 +249,12 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
         per_device.plot_promql(
             PlotOpts::gauge("Used (Per-GPU)", "mem-used-per-gpu", Unit::Bytes)
                 .with_row_label("GPU"),
-            "sum by (id) (gpu_memory{state=\"used\"})".to_string(),
+            "sum by (id, vendor) (gpu_memory{state=\"used\"})".to_string(),
         );
         per_device.plot_promql(
             PlotOpts::gauge("Free (Per-GPU)", "mem-free-per-gpu", Unit::Bytes)
                 .with_row_label("GPU"),
-            "sum by (id) (gpu_memory{state=\"free\"})".to_string(),
+            "sum by (id, vendor) (gpu_memory{state=\"free\"})".to_string(),
         );
     }
 
@@ -278,7 +278,7 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
             )
             .percentage_range()
             .with_row_label("GPU"),
-            "sum by (id) (gpu_dram_bandwidth_utilization) / 100".to_string(),
+            "sum by (id, vendor) (gpu_dram_bandwidth_utilization) / 100".to_string(),
         );
     } else {
         dram_bw.plot_promql_full(
@@ -352,7 +352,7 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
             PlotOpts::gauge("Power (Per-GPU)", "power-watts-per-gpu", Unit::Count)
                 .with_axis_label("Watts")
                 .with_row_label("GPU"),
-            "sum by (id) (gpu_power_usage) / 1000".to_string(),
+            "sum by (id, vendor) (gpu_power_usage) / 1000".to_string(),
         );
     } else {
         draw.plot_promql_full(
@@ -383,7 +383,7 @@ pub fn generate(data: &dyn MetricsSource, sections: Vec<Section>) -> View {
             PlotOpts::gauge("Temperature (Per-GPU)", "temp-per-gpu", Unit::Count)
                 .with_axis_label("°C")
                 .with_row_label("GPU"),
-            "max by (id) (gpu_temperature)".to_string(),
+            "max by (id, vendor) (gpu_temperature)".to_string(),
         );
         temps.plot_promql(
             PlotOpts::gauge("Max (°C)", "temp-max", Unit::Count).with_axis_label("°C"),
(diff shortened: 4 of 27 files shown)


yangxi:
Re-verified on a third host now that it is back up.

**hyb** (Alder Lake UHD 730 + RTX 5080) — the mixed-vendor case this PR is
written for. Builds clean, 747 tests pass, clippy clean. Both samplers come up
healthy, and a live snapshot carries two GPUs from different vendors at the same
id:

```
vendor=intel   id=0
vendor=nvidia  id=0
```

That is the collision the `(id, vendor)` grouping disambiguates: `sum by (id)`
would have summed these two devices into one line, and the heatmap indexed both
onto row 0 with one overwriting the other.

All three hosts are now green on the final branch — sky (Arc A770 + UHD 630),
hyb (above), and amd (R9700 + Raphael iGPU).

One caveat from the description still stands, and this run does not close it:
the *query-level* conflation needs a metric name published by both vendors, and
hyb's Intel GPU is integrated, so it publishes no `gpu_memory` /
`gpu_utilization` / `gpu_temperature` to collide with NVIDIA's. Reproducing
that end to end needs a discrete Intel card alongside another vendor, which no
host here has. It remains covered by unit tests only.

yangxi:
Correction to the last bullet in the description: the three WASM test failures
are **not** an existing problem on `main` — they were a stale local build
artifact, and they pass once it is rebuilt.

`site/viewer/pkg/` is gitignored, and mine dated to 19 June. The tests load the
compiled `.wasm` directly, so they were exercising a three-month-old binary that
predated both features they assert on — `classify_sources` (`lib.rs:294`) and
`timestamps()` (`lib.rs:839`) are both present in the source.

After `./crates/viewer/build.sh`:

```
tests 273
pass  265
fail  0
```

Nothing to commit — `pkg/` is a build artifact and CI builds it fresh. Anyone
reproducing locally needs the same rebuild before those three tests are
meaningful.

brayniac (changes requested):
Reviewed in a worktree with the branch merged onto current main. It merges cleanly, `cargo check --workspace --all-targets` is clean, and **690** Rust plus **262** JS tests pass on the merged tree.

The sampler itself is in good shape and I want to say so before the findings: overhead is genuinely *measured* (6.4 us per PMU group read, 5.4 us per DRM ioctl), the read is off the scrape cadence via `spawn_blocking` on its own interval with the principle-17 departure documented, engine members use an explicit sparse set because the `MAX_ENGINES` blocks are sparse, the grouping audit is complete (17 `(id, vendor)`, zero bare `by (id)` left), and the systeminfo `if gpus.is_empty()` bug is real and correctly fixed. That is a higher bar than most samplers clear.

Three blocking items below, then five smaller ones.

## 1. `resolveStyle` is changed globally, and it is not a GPU-only change

`src/viewer/assets/lib/charts/metric_types.js:57`

I ran the pure function on both branches:

```
                                  main       PR
single entity, has id             line   ->  heatmap
two entities, has id              heatmap    heatmap
```

`resolveStyle` governs every gauge/delta_counter plot — `data.js:383`, `data.js:693`, `chart.js:826`, `chart.js:1112` — not just GPU. So on any host with exactly one of an entity, per-entity charts flip from a line to a one-row heatmap: a 1-vCPU VM, a single NIC, a single disk, one cgroup. A one-row heatmap encodes value as colour and drops the y-axis, so for a single series it is strictly less readable than the line it replaces. The function's own doc comment ("Single series -> 'line'") is now stale.

There are **no tests for `resolveStyle` anywhere in the repo**, and this PR adds none — while adding thirteen for the GPU selector.

`data.js:693` already honours `plot.opts.style || resolveStyle(...)`, but nothing ever sets `opts.style` — it is a dormant override hook. Adding a `style` field to `PlotOpts` and setting it on `gpu.rs`'s per-GPU plots gets the intended GPU behaviour without changing chart type for every other subsystem.

## 2. Every per-CPU heatmap loses its axis title and tooltip prefix

`src/viewer/assets/lib/data.js:719` + `src/viewer/assets/lib/charts/heatmap.js:86`

`plot.row_labels = rowLabels` is assigned unconditionally, and `promqlResultToHeatmapTriples` always builds a non-empty array — for a CPU heatmap it is `["0","1","2",...]`, because `m.id` is set and `m.vendor` is not. `heatmap.js` treats a truthy `suppliedRowLabels` as "these rows name themselves" and drops `name: yAxisLabel || 'CPU'` from the y-axis, and renders tooltips as `String(rowLabels[cpu])`.

Run against a plain CPU-shaped result, no GPUs involved:

```
rowLabels for a CPU heatmap: ["0","1","2"]
truthy (heatmap.js suppliedRowLabels)?  true
  yAxis name  -> DROPPED (no "CPU" title)
  tooltip row -> "2"  (was "CPU 2")
```

So on *every* host, the per-CPU / softirq / scheduler heatmaps lose the "CPU" y-axis title and their tooltips read `2` instead of `CPU 2`. The code's own comment states the opposite intent — "Numeric rows keep the name, since '0' alone does not say what it counts" — so this is a bug against stated design, not a judgement call. Compare mode still shows them (`viewer_core.js:167` discards `rowLabels`), so the two paths now disagree.

Gate on a label actually differing from its row index — e.g. only set `row_labels` when some series carries a `vendor`.

## 3. `set_member_bound(gpus.len())` assumes dense ids, but ids come from the PMU list

`src/agent/samplers/gpu/linux/intel/mod.rs:242`

`init` assigns `id` from `pmus.iter().enumerate()` and pushes only the GPUs whose `Gpu::new` succeeded. If an earlier PMU fails while a later one succeeds, the survivor has `id == 1` while `gpus.len() == 1` — so the device group declares member `{0}` only.

The snapshot walk iterates `members(member_set, member_bound, entries)`, so `gpu_frequency_sample{id="1"}` and `gpu_memory{id="1"}` are dropped entirely, and index 0 is walked but never written. For the `CounterGroup` frequency metrics that publishes `Some(0)` — a phantom all-zero series for a GPU that does not exist — which is the exact failure the two-group split exists to prevent.

The engine group is unaffected, because it builds an explicit set from `gpu.id`. That asymmetry is what makes this easy to miss. `set_member_set` over the surviving ids fixes it.

This compounds with finding 4: an xe-driver GPU alongside an i915 one is precisely the partial-failure shape.

## 4. `xe` is matched but only i915 event names are recognised, and the failure blames permissions

`src/agent/samplers/gpu/linux/intel/pmu.rs:117`

`for driver in ["i915", "xe"]` matches both, and the module doc says so explicitly. But `discover_engines` looks for `<engine>-busy` and the globals for `actual-frequency` / `requested-frequency`, which are i915's vocabulary; every test fixture in `pmu.rs` uses i915 names too. My understanding is that `xe` exposes `engine-active-ticks` / `engine-total-ticks` and `gt-actual-frequency` instead, with engine identity in `config` bits rather than the event name — please confirm, since you have the hardware and I do not.

If that is right, on an xe host every `Gpu::new` returns `NotFound`, `gpus.is_empty()` fires, and the sampler reports:

> `Intel GPU present but no PMU counters could be opened (needs CAP_PERFMON or kernel.perf_event_paranoid <= 2)`

which blames permissions for a driver mismatch, on hardware where `xe` is the default (Lunar Lake, Battlemage). Either handle xe's vocabulary or stop matching the `xe` prefix and say which drivers are supported.

## 5. Smaller

- **`render_node()` aborts the whole scan on one bad node.** `crates/systeminfo/src/hwinfo/gpu.rs:830` — `canonicalize(...).ok()?` sits inside the `for` loop, so one `renderD*` whose `device` link will not resolve returns `None` for the entire function instead of continuing. `readdir` order is arbitrary, so on a host with an extra or virtual render node the real card may never be reached. The sibling in `src/agent/samplers/gpu/linux/intel/drm_memory.rs` correctly uses `continue`.
- **"integrated" is shown for memory that is merely unknown.** `section_views.js:113` — `gpu.memory_bytes ? ... : 'integrated'`. The NVIDIA `/proc` fallback this PR adds always sets `memory_bytes: None`, so on a host with the driver but no NVML every discrete NVIDIA card is now labelled "integrated" in System Info.
- **Six of the thirteen new tests exercise local copies, not shipped code.** `tests/gpu_selection.test.mjs:96,102` re-implement `seriesName` and `heatmapRowLabels` rather than importing them, so a regression in `data.js` would not fail them. They have already diverged: the local `seriesName` returns `null` with no vendor, where `data.js` falls back to the first non-`__name__` label. `promqlResultToHeatmapTriples` is exported — importing it is what would have caught finding 2.
- **The Intel PMU dashboard group is emitted unconditionally.** The per-engine subgroups are gated on `label_values(...)`, but the Frequency subgroup and the final `view.group(pmu)` are not, so an NVIDIA- or AMD-only recording renders an empty "Intel GPU Performance Counters" section. (`amd_pmu` has the same pre-existing shape, so this doubles it.)
- **Metadata may be inserted for events that were never opened** — the `GLOBAL_EVENTS` and per-engine loops appear to tag `id` regardless of whether the unit check or `build_with_group` succeeded, which would put permanently-null series in the declared member set. I did not verify this one as carefully as the others; worth a look.

## Not blocking, for the record

The mixed-vendor conflation having no end-to-end hardware proof is fine — it is stated plainly and the unit tests around it are meaningful. I could not render-check the GPU charts either: there is no GPU-bearing parquet fixture in the repo, and the new `viewer-render` harness needs real data. A recording from sky or hyb would settle findings 1 and 2 visually in about a minute.

https://claude.ai/code/session_01KheDzaD5bnVWcmdocmk2eW

yangxi:
Thanks — this was a careful review and every finding held up. I reproduced each
one before fixing it, including #5's last bullet, which you flagged as
unverified: metadata is tagged from `discover_engines`, never filtered by what
opened, so it is real.

Pushed as `435100bf`.

## 1 — `resolveStyle` (blocking)

You were right that this was not a GPU-only change, and your suggested fix is
better than mine. Reverted to the original inference. `PlotOpts::style` is now
a real builder and is set on the eleven per-GPU plots, so the GPU behaviour
arrives without touching any other subsystem.

Verified against a full 1,717-column capture from hyb (all samplers enabled):

```
cpu   Busy % (Per-CPU)      n=16  style=(infer)  -> heatmap
gpu   GPU % (Per-GPU)       n=1   style=heatmap  -> heatmap
```

Every CPU chart still resolves by inference and is unchanged; only the GPU
plots carry the override. Added the five `resolveStyle` tests the function was
missing.

## 2 — per-CPU heatmap axis title (blocking)

Confirmed exactly as you described. `row_labels` is now withheld when every
label equals its own row index, so the array is `null` for a CPU heatmap and
the renderer keeps its title and `CPU 2` tooltips. Verified on the same capture
with 16 real CPUs:

```
series: 16  first rows: ['0','1','2','3']
informative (would override axis title): False
-> CPU axis title KEPT
```

## 3 — `set_member_bound(gpus.len())` (blocking)

Correct, and my comment asserting the ids were dense was simply wrong. Both
groups now use explicit member sets, and membership comes from the indices
whose counters actually opened — which also fixes #5's last bullet, since the
same `live_indices` set gates the identity metadata.

## 4 — `xe`

I cannot confirm your reading of xe's vocabulary: none of the three hosts has
xe hardware, so I would be guessing. Taking the second option you offered — the
`xe` prefix is no longer matched, and the doc comment records what support
would require. An xe host now reports `Unsupported` with an accurate reason
instead of blaming permissions. Stale `i915/xe` claims removed across the tree.

## 5 — smaller

All four addressed: `render_node` continues past an unresolvable node;
absent memory reports `architecture` rather than asserting "integrated" (your
point about the `/proc` fallback making that wrong for discrete cards was
exactly right); the Intel group returns early on a recording with no Intel
metrics; metadata is gated on `live_indices`.

On the tests re-implementing rather than importing — that one stung, because
importing `promqlResultToHeatmapTriples` is precisely what would have caught
finding 2. They now import it, with a CPU-shaped case among them.

## Fixture

Good call on the GPU parquet. The spec now defaults to a full capture — every
sampler, not just the GPU one — since these viewer changes are not GPU-specific
and an isolated GPU recording cannot see a CPU-chart regression at all. That is
what produced the numbers above. Happy to attach a recording if a fixture in
the repo would help; say the word on where it should live.

730 Rust tests on sky, hyb and amd; 277 JS tests; clippy clean.

yangxi:
Full-capture verification on all three hosts, with every sampler enabled — the
spec now defaults to that, since these viewer changes are not GPU-specific and
an isolated GPU recording cannot see a CPU-chart regression.

Recordings are 1717 / 1716 / ~2000 columns across 27 samplers, versus 29 for the
GPU-only captures.

**Finding 1** — CPU charts resolve by inference and are unchanged; only the
eleven per-GPU plots carry the override:

```
             CPU (inferred)                GPU (explicit)
sky  (12c)   Busy % (Per-CPU)  n=12  ->heatmap    Compute Busy %  n=1  style=heatmap
hyb  (16c)   Busy % (Per-CPU)  n=16  ->heatmap    GPU %           n=1  style=heatmap
amd  (24c)   Busy % (Per-CPU)  n=24  ->heatmap    GPU %           n=2  style=heatmap
```

**Finding 2** — on hyb's 16 real CPUs the labels come back `['0','1','2','3'...]`,
uninformative, so the array is withheld and the axis keeps its "CPU" title.

GPU labels still correct on every host: sky `intel 1`, hyb `nvidia 0`, amd
`amd 0` / `amd 1`.

brayniac (approved):
Re-reviewed against the branch merged onto current `main`. Every finding from the previous round verified **in the code**, not from the commit message.

| finding | verified |
|---|---|
| 1. `resolveStyle` changed globally | Reverted to `result.length > 1`, so a single entity stays a line. `PlotOpts::with_style` is a real builder, set on exactly 11 per-GPU plots, and `data.js:706` honours it ahead of the inference. Five tests added in `tests/resolve_style.test.mjs` — the function previously had none anywhere in the repo |
| 2. per-CPU heatmaps lose axis title + tooltip prefix | `rowLabelsAreInformative` withholds the array when every label merely restates its row index. Checked the sparse-id case too: `Array.prototype.some` skips holes, so sparse GPU ids behave correctly rather than accidentally |
| 3. `set_member_bound(gpus.len())` assumes dense ids | Both groups now build explicit sets — `device_members` from surviving `gpu.id`, `engine_members` gated on `live_indices` |
| 4. `xe` matched but only i915 vocabulary recognised | `parse_pmu_name("xe") → None`, pinned by a test. `systeminfo` still matches `i915 \| xe`, which is correct and not an oversight: the inventory should list an xe GPU even though the PMU sampler does not sample it |
| 5. five smaller | all five: `render_node` uses `continue`; absent memory reports `architecture` rather than asserting "integrated"; the Intel dashboard group returns early; the tests import `promqlResultToHeatmapTriples` instead of re-implementing it |
| 6. metadata for unopened events (I flagged this unverified) | fixed for engines via the `live_indices` gate |

The CodeQL clippy finding (`for loop over a single element`, pmu.rs:144) was the `for driver in ["i915", "xe"]` loop — resolved as a side effect of finding 4. `cargo clippy --all-targets` clean.

Merges cleanly onto `main`; 702 Rust + 17 integration + 292 JS pass on the merged tree.

I did not re-measure the sampler — no Intel GPU here — so the 6.4 us / 5.4 us figures stand as reported in the original round.

## Non-blocking, filed as follow-ups rather than held here

- `GLOBAL_EVENTS` identity metadata and `device_members` are pushed unconditionally while engine metadata is gated on `live_indices`. I traced it and it is harmless today: `refresh` writes only live counters, an unwritten slot keeps the `UNWRITTEN` sentinel, and exposition skips `None`, so the series is absent rather than a phantom zero. Worth aligning with the rule the commit itself states.
- `crates/dashboard/src/dashboard/gpu.rs` has no generator test naming the new metrics. Pre-existing — `gpu.rs` had no `cfg(test)` module on `main` either — so not a regression from this PR.
- Findings 1 and 2 were never render-checked visually; there is still no GPU-bearing fixture in the repo.

PR #1201, opened 2026-09-12
feat(rez): format hardening — identity, version gate, reader atomicity

brayniac:
## Summary

A critical review of `crates/rez` found the container design sound and the *format* layer thin. This PR lands the review as a journal entry (`docs/journal/2026-09-11-rez-format-hardening.md`) and then works through its seven items, one commit each:

1. **Version gate.** `RezDb::create` stamps `application_id` (`REZ\0`) and `user_version` into the SQLite header; `open`/`open_bytes` refuse a newer schema by name, a foreign database as not-a-rez, and a catalog-less copy with the `recording snapshot` hint, through a typed `OpenError`. Pre-stamp archives still open. Detection reads the id from the 100-byte header.
2. **Reader atomicity.** Segments and the live WAL are read in one `BEGIN DEFERRED` snapshot, and so is the whole open-time catalog probe. New test seals from a second connection between the two reads; on the old form it reads **0 of 3 rows** (negative control run before committing).
3. **Recording identity.** `recordings.uuid` (v4, minted from SQLite's `randomblob` so the wasm reader needs no random source), preserved by every copy, printed by `recording metadata`. `combine` refuses equal uuids outright and identical label sets unless `--allow-duplicate-labels`.
4. **Producer epoch.** The agent mints one id per process and carries it in every snapshot's metadata as `producer_epoch`; the writer records it and any change (`producer_epochs` history plus a `producer_epoch` timeline event) as it happens, through a new `UpdateMetadata` writer message. Two recordings with one epoch over overlapping time are observations of one series, not two to sum.
5. **Specification.** `docs/rez-format.md`: container, catalog, table-key rule, live-WAL predicate, segment and WAL encodings, reserved keys, time model, retention, copies, compatibility rule. Stale tar-era and "Prometheus forces parquet" text fixed in `rez.rs`, `docs/parquet_metadata.md`, README, `record --help`, CLAUDE.md.
6. **Typed container errors.** `RezDb`/`RezTx` return `DbError` carrying SQLite's code; the writer loop retries the retryable class (3 attempts, ~310 ms), drops a still-failing tick and stops after 30 consecutive drops, defers a failing seal to the next batch (`seq` now advances only after commit), and re-commits a constraint-failing tick per recording so only the colliding recording loses rows. Both new tests fail on the old single-`?` commit path.
7. **Reopen for append.** `RezArchive::open` + `resume_recording`: `seq` and clock offsets continue from the catalog, a backwards anchor or tick is refused, the session is recorded under `writer_sessions` with a `writer_session` event, `complete` is cleared and re-set by the session's finalize. Hindsight wiring is deferred to the backlog with the primitive in place.

Deferred findings (retention anchor loss, first-segment-only routing probe, read-only opens, `lib.rs` facade, metriken-free `TableWriter`, the smaller items) are in `docs/backlog.md` under "`.rez` — format hardening".

## Test plan
- [x] `cargo test -p rez --features test-support` — 208 pass (190 → 208)
- [x] `cargo test -p rezolus` — 694 pass + integration suites
- [x] `cargo test -p viewer -p report-save -p dashboard` — 14 / 17 / 85 pass
- [x] `cargo clippy --workspace --all-targets` clean; `cargo fmt --check` clean
- [x] `cargo check -p rez --no-default-features --target wasm32-unknown-unknown`
- [x] `tests/help_text.rs` after the `record --help` change
- [x] Negative controls run for items 2 and 6 (tests fail on the pre-change code paths)

https://claude.ai/code/session_0112Ajcj2SXjQcYjrPiYHqDp


brayniac:
Closing for now: this landed against rezolus's crates/rez, but the .rez format has moved to iopsystems/dendro and the review is being redone there. The branch stays for reference; the version gate and read-seam commits are the two with standalone value if crates/rez outlives the cut-over.

Issue #1185, opened 2026-09-10
Viewer: no TCP dashboard section — six default-collected TCP metrics are never plotted


brayniac:
## Summary

Four TCP samplers are enabled by default in `config/agent.toml` and collect on
every host. Between them they produce six metrics that **no dashboard panel
ever draws**. There is no `tcp.rs` in `crates/dashboard/src/dashboard/` — the
only TCP content anywhere is a one-plot "TCP" group inside `network.rs:52-70`
(packet latency) and retransmits in `exceptions.rs`.

Found while auditing all 120 sampler-defined metric names against every PromQL
query in `crates/dashboard/src/` and `src/viewer/tui/`. Full audit table at the
bottom of this issue.

## Collected on every host, never plotted

| Metric | Sampler | Description as shipped |
|---|---|---|
| `tcp_srtt` | `tcp_receive` | Distribution of TCP smoothed round-trip time |
| `tcp_jitter` | `tcp_receive` | Distribution of TCP latency jitter |
| `tcp_connect_latency` | `tcp_connect_latency` | Distribution of latency for establishing outbound connections (active open) |
| `tcp_bytes` | `tcp_traffic` | TX/RX bytes |
| `tcp_packets` | `tcp_traffic` | TX/RX packets |
| `tcp_size` | `tcp_traffic` | Distribution of the size of TCP packets transferred |

`tcp_srtt` and `tcp_connect_latency` are close to the most directly useful
things Rezolus collects for "is the network the problem, or is it us" — and
neither is visible without hand-writing a query in the Query Explorer.

`tcp_bytes` is also the one metric that would independently corroborate the
stacked-interface double count in #1184, since it counts at the socket layer
rather than per-netdev. No panel pairs them.

## Also unplotted in `network_interfaces`

- `network_transmit_complete` — "Packets successfully transmitted by the
  driver." This is the denominator that makes the transmit-timeout rate in
  `exceptions.rs` mean something. Right now the timeout panel tells you
  failures with nothing to scale them against.
- `network_transmit_busy` — "Packets encountering retryable device busy
  status. High rates indicate transmit path backpressure."

Small related bug: `network_transmit_complete`'s own description
(`src/agent/samplers/network/linux/interfaces/stats.rs:60`) says "Compare with
`network_transmit_packets`" — that metric does not exist anywhere in the tree.
The real one is `network_packets{direction="transmit"}`.

## Suggested shape

[...]

Issue #133, opened 2023-12-12
Filesystem occupancy metrics


brayniac:
We've received a request to add disk usage metrics to Rezolus, particularly free space on filesystems. This issue is to track adding that

bxffour:
Hi there,

I'm Nana, and I'm interested in contributing to this issue. Before starting, I'd like to understand the expected output format and requirements for these metrics.

Could you please provide more details about what you're looking for in terms of the metrics representation?

Thanks!

brayniac:
Hi @bxffour - thank you for your interest in contributing. This issue probably could've used a little more detail and I think I probably shouldn't have used the word 'utilization' in the issue description. What was wanted here was actually metrics around free space on filesystems. I've adjusted this issue to hopefully clarify the original intent.

Since that request for support, I've been giving more thought to where Rezolus provides value compared to other telemetry agents. I think the short answer there is that the value comes from low overhead collection of utilization and performance metrics. I feel like this request was more about health metrics, which I feel other projects tend to cover pretty well. 

With that framing in mind, I'm not sure this particular issue should remain open. I think efforts would be better spent in building out features which differentiate Rezolus from other telemetry agents and provide unique value.

If you'd like to chat more about ideas for the project and areas to contribute, feel free to join our discord (linked in the readme) and ping me to discuss.

brayniac:
Closing this one out for now. Willing to revisit if there's sufficient demand.

PR #1204, opened 2026-09-12
feat(filesystem): local filesystem occupancy sampler + viewer section

### LINKED ISSUE #1202: No filesystem free-space metric: a filling mount is invisible in a recording
### Summary

Rezolus has no metric for how full a mounted filesystem is. `blockio` reports operations, bytes and latency per block device, and `drivehealth` reports drive temperature and NVMe thermal-throttle counters, but nothing reports free or total space per mount. A host that is about to run out of disk looks healthy in a Rezolus recording right up to the write that fails.

This revisits #133, closed in 2024 with "willing to revisit if there's sufficient demand."

### Expected behavior

Free and total space for each mounted filesystem, excluding pseudo-filesystems such as tmpfs, devtmpfs and overlay, is available as a Rezolus metric with the mount as a label, so a recording can show a filesystem filling and a dashboard can alert on it.

### Actual behavior

No filesystem occupancy metric exists in any sampler. Operators who need it run a second agent or `df` beside Rezolus.

### Environment

- Linux, x86_64 and aarch64 hosts; applies to every kernel version Rezolus supports, since the data comes from mount tables rather than from BPF.

_Generated by Claude on Yao's behalf._

thinkingfish:
Problem

Rezolus has no metric for how full a mounted filesystem is. `blockio` reports the device and `drivehealth` the drive; nothing reports the mount, so a host about to run out of disk looks healthy in a recording until the write that fails. #133 asked for this in 2023 and closed on overhead versus value, "willing to revisit if there's sufficient demand"; #1202 is that demand, with the two things that changed since: per-sampler intervals exist, and `drivehealth` established the throttled off-cycle sweep pattern for slow health gauges.

Solution

A `filesystem` sampler (`src/agent/samplers/filesystem/`) that, once per `[samplers.filesystem] interval` (default 60s, off the scrape cycle via `spawn_blocking`), reads `/proc/self/mountinfo`, classifies every mount, and issues one `statvfs` per **local** filesystem. Six gauges per filesystem (one series per superblock), labeled `mount`, `fstype`, `devnum` (major:minor) and, when block-backed, `block_device` (the kernel's partition or mapped-device name, which does not join `drivehealth`'s drive name): `filesystem_total`, `filesystem_free`, `filesystem_available` (bytes), `filesystem_inodes_total`, `filesystem_inodes_free`, and `filesystem_readonly` (0/1, read-only as a whole: superblock `ro`, or ext4 `emergency_ro`, which current kernels set after an error instead of the superblock flag; a read-only bind of a writable filesystem reads 0, a full filesystem stays writable, and an XFS shutdown shows neither). One acquisition group for the sweep (principle 18's device-sweep shape, as `drivehealth`).

**Local mounts only, by construction.** Classification runs before any `statvfs`: block-backed types with a `/dev` source plus `zfs` are sampled; network types (`nfs`, `cifs`, `ceph`, ...), every FUSE type, autofs triggers and pseudo-filesystems are never touched. A `statvfs` on a `hard` network mount blocks until the server answers with no timeout the agent can set, and one on an autofs trigger starts a mount. Network mounts stay out of scope until asked for; the decision and the single opt-in TODO live in the sampler's module doc, per the discussion on #1202. Only mounts that path lookup reaches are sampled, resolved from the root mount through each stack of mounts, since `statvfs` on a covered mount's path reaches whatever is stacked on top; each read is refused unless its descriptor's mount id matches the table's. Found in local review, fixed in 44c0f8cd, 8c5af75c and 85d51be8. Since faf1222f a chroot table whose root mount is omitted samples nothing, because that mount's type is unknown and may be NFS; skipped candidates are logged as warnings with a reason when the reasons change.

The mount table is rescanned every sweep rather than enumerated once at startup, because a filesystem mounted after the agent started and then filling up is the case the metric exists for. Slots are stable per `major:minor` while mounted; an unmounted filesystem's values are unset and labels cleared, and the slot is reused.

Also: a Filesystem dashboard section, including a read-only card (`sum by (mount)` per gauge, so the viewer draws one line per filesystem and filesystems that come and go show up on the next render), `config/agent.toml`, `docs/metrics.md`, the analysis-side sampler registry, a journal entry with the measurements, and the backlog items.

**Per-sweep member bound, ruled acceptable.** The sweep task updates the acquisition group's member bound each sweep (one past the highest occupied slot), so a 3-mount host is walked as 3 members rather than `MAX_MOUNTS`. Every other sampler sets its bound only during init, including both GPU samplers, which set 0 and then the device count. Principle 18 in `docs/principles.md` now allows a changing population to revise its bound on every read, from the group's single writer, before stamping, and names the accepted costs: a one-snapshot incoherence (the builder reads the bound, values and labels per metric family without locking out the writer) and a schema change per membership change. Rejected: a fixed bound of 64 (empty rows in every snapshot), membership by labels (a builder change for the same result), and a bound frozen at startup (later mounts invisible).

Result

Measured per the reviewing-samplers recipe (unprivileged, only this sampler enabled, 3 local filesystems):

| build | `refresh()` on the scrape path | full sweep, off-cycle |
|---|---|---|
| debug | 3–22 µs | 626–1139 µs |
| release | 0–8 µs (first call 91 µs) | 330–570 µs (read 220–350, parse 100–200, statvfs+publish 11–50) |

The sweep is a cold blocking-pool thread once per interval; the same work in a hot test loop is about 100 µs, so the cold number is the one reported. Amortized at the 60 s default: about 10 µs per second, 0.001 % of one core.

Also in this PR, at Yao's request: the principles checklist and the reviewing-samplers skill gain a **"you record it, you display it"** item — a metric a sampler adds is plotted in the viewer in the same change, with a generator test naming it. #1185 is the standing example of what happens otherwise.

**After round 11 and brayniac's review** (faf1222f): a slot whose identity changes starts without its old reading; a panic restart clears every slot; stuck sweeps warn after three intervals and no longer advance the throttle, and the first sweep no longer runs inline at startup; `MAX_MOUNTS` is 256 with the over-cap warning on change. Follow-ups: reading the member bound once per snapshot (#1207), a runtime degraded sampler status (#1208), and ownership of the site's version label (#1209).

**Filesystem context, tracked in #1206:** source, mount root, and per-mount and superblock option strings are not recorded in this PR.

**Known limitation, tracked in #1205:** a `.rez` recording keeps a slot's first labels when the slot is relabeled or reused within a segment. The cause is in the `crates/rez` group table builder and `cpu_usage`'s per-PID task slots share it, so it is not fixed in this PR.

**Not measured, and tracked:** sweep cost on a many-thousand-mount container host. Measurements so far come from one host with an 87-line mount table and 3 local filesystems; the kernel generating the table dominates the sweep and grows with it. The journal's Deferred list and `docs/backlog.md` carry the reopen condition: measure on such a host before enabling fleet-wide.

**Also fixed after review-skill testing** (05724847): a local mount below an NFS, FUSE or autofs mount is no longer sampled, since its lookup walks the blocking mount; a panicking sweep no longer stops the sampler for good; inode gauges are absent on filesystems with no inode limit (btrfs, vfat), so *Inodes Free %* no longer divides 0 by 0, and *Used %* now matches `df`; a failed first mount-table read no longer leaves the group unbounded; a sweep with no successful read keeps the previous values with the previous window. The `reviewing-samplers` skill and principles checklist gain a **series context** item (14f45439).

Tests: 24 on the mount-table parser and classifier (visibility cases, skip reasons, and the superblock read-only flag), 21 on slots and relabeling, `statvfs`, its mount-id refusal and a regular-file read, vacate/label, absent inode gauges, the panic latch and restart, the failed-read bound, publication when no read succeeds, identity changes, `block_device` naming, warn-on-change, the held-latch refresh, and the end-to-end sweep against a readable local mount, 4 on the dashboard section; the registry guard tests pass unchanged.

Closes #1202. Revisits #133 — thanks to @bxffour, who offered to build this back then, and to @brayniac for relaying the original request and for the drivehealth pattern this follows.


brayniac:
Early feedback while this is still draft — read the whole branch (11 commits, +1816/−34), didn't run it on a Linux host.

Short version: the sampler-side engineering is unusually careful. Classify-before-lookup, mount-id verification on an `O_PATH` fd, visibility resolved from the root rather than by string prefix, `emergency_ro` — that last one is a kernel detail most implementations of this metric get wrong. The mount-table tests are real tests with mutation checks behind them. It's close. Here's what I'd settle before it leaves draft.

## 1. The `device` label means something different here than everywhere else

`label()` publishes `device = mount.device`, which is `"8:1"` — major:minor. But `drivehealth` publishes `device` as the kernel name (`nvme0`, `sda`), and `docs/metrics.md:221` documents it that way. So:

- `filesystem_available / on(device) drive_temperature` silently returns nothing, which is the natural first thing an operator tries.
- `MountEntry.source` (`/dev/sda1`) is parsed, carried through the whole pipeline, and never published — it's the one field a human would recognize on sight.

Suggestion: resolve the kernel name once per filesystem per sweep via `readlink /sys/dev/block/<major>:<minor>` and basename it. That's the name `blockio` and `drivehealth` both use, and it gets dm/LVM right (`/dev/mapper/root` → `dm-0`, which basenaming `source` would not). Keep major:minor if you want it, under a different key. ZFS has no block device, so it'd just be absent there.

Cheap to change now, expensive once the label ships.

## 2. Resolution failure is total, silent data loss

`mounts::visible()` returns `Vec::new()` on two paths — `starts` not being exactly one, and `climb` hitting two mounts at the same `(parent, point)`. `sweep()` then publishes 0, calls `guard.discard()`, and logs only at `debug!`. A host where resolution bails looks *exactly* like a host with no local filesystems.

For a sampler whose entire reason to exist is catching a disk about to fill, that's the worst failure mode available. `two_omitted_parents_sample_nothing` and `an_ambiguous_attachment_samples_nothing_at_or_below_it` both enshrine it as intended, which is a fine policy — but it needs a `warn!` naming the reason, fired when the table had local candidates and resolution returned none. Ideally `visible()` returns a reason rather than an empty vec, so `rezolus status` can tell "no local filesystems" from "gave up".

## 3. The `set_member_bound` question

Bless the revision, fix the thing underneath it.

The revision itself is right. Pinning to `MAX_MOUNTS` would emit 61 empty slots per tick forever, both GPU samplers already revise post-init, and the rewritten `timing.rs` doc describes the tree accurately.

The hazard isn't the revision, it's the read side. `snapshot.rs:1666` reads `member_bound()` **inside the per-metric loop**, so the six gauge families each read it independently. A sweep landing mid-walk gives one group table ragged membership across families in a single tick — `filesystem_total` at 5 members, `filesystem_readonly` at 3. That's latent for the GPU samplers, whose population changes once at discovery; this is the first sampler where it changes routinely.

The contained fix is in `snapshot.rs`: read `member_bound`/`member_set` once at the group's first touch this tick and cache them on the group entry, instead of per metric. Then a sweep lands wholly before or wholly after a walk, and per-sweep revision is unambiguously safe rather than "safe enough at 60s". Probably its own PR — but that's the answer to the question you asked, and I'd rather it be a known follow-up than an unstated assumption.

## 4. A parked sweep is indistinguishable from a healthy one

The module doc accepts that a network mount stacked between the table read and the `open` can park the sweep thread. If that happens: `reading` stays `true` forever, `refresh()` keeps bumping `last_read` every 60s and bailing on the latch, and the six gauges hold their last values indefinitely. The recording then shows a filesystem that stopped changing — which, for an occupancy metric, reads as "idle disk", not "broken sampler".

Two cheap mitigations: bail *before* bumping `last_read` when the latch is already held (drivehealth has the same ordering, but its reads are bounded and this one isn't), and warn when `reading` has been set for more than a few intervals — that's what makes it visible to `rezolus status`.

Separately, the startup sweep in `Filesystem::new` is synchronous and inline, so the same race hangs agent startup rather than one sweep. Since `member_bound(0)` is already meaningful, routing the first sweep through `spawn_blocking` too costs one empty snapshot and removes that.

## 5. "Used %" doesn't agree with `df`

`crates/dashboard/src/dashboard/filesystem.rs` computes `1 - available/total`. `df`'s Use% is `used/(used+available)`. On a default ext4 with the 5% superuser reserve, an empty filesystem charts as 5% used while `df` says 0%.

`docs/metrics.md` anchors the whole section on "the number `df` reports as available", so a Use% that disagrees with `df -h` is the kind of mismatch that makes people stop trusting the panel. `gpu.rs:189` already uses the `used/(used+free)` shape, so there's house precedent: `(total-free) / ((total-free) + available)`. Or keep the current formula and rename the card to what it measures — headroom consumed, reserve included.

## 6. The inode ratio divides by a number the card documents as zero

`sum by (mount) (filesystem_inodes_free) / sum by (mount) (filesystem_inodes_total)` — and the card's own description says "btrfs and vfat report zero". So any btrfs or vfat mount renders that series as `+Inf` or `NaN` under `.percentage_range()`. Worth checking what the chart actually does with that before it ships; the fix is filtering the denominator.

Minor, same file: `filesystem_inodes_total` is only ever a denominator, never plotted. That satisfies the letter of the new checklist item ("named in a query") but not quite its spirit.

## 7. The two version bumps, which are not the same thing

The `Cargo.toml`/`Cargo.lock` bump to `5.20.1-alpha.1` is intended — no issue.

The `site/docs/*.html` bump `v5.19.1` → `v5.20.0` is the *correct value*: v5.20.0 is the latest published release and the site was stale. So don't revert it. But it doesn't belong to this PR either, and the fact that it drifted at all is the actual finding: `53ed5244` ("release: prepare v5.20.0") doesn't touch `site/docs/` — nothing in the release flow does. The last thing to update that string was #1191, a CI PR that happened to notice. The string is owned by nobody, so it's correct only by coincidence, and right now it's correct in a branch that hasn't merged.

Worth a separate issue: either the release process updates it, or the page reads the version from one source instead of six hand-edited copies.

## Smaller notes

- `MAX_MOUNTS = 64` with the over-cap `warn!` firing **every sweep** — a log line per minute, forever, on a host that exceeds it. Warn on change instead. Also 64 feels low for a ZFS host: each dataset is its own mount with its own anonymous device id, so a pool with 80 datasets silently loses 16.
- `docs/metrics.md` says local filesystems "answer from in-memory superblock counters and issue no I/O". True for ext4/XFS; btrfs `statfs` does real space accounting under locks. Not wrong enough to matter at 60s, but the blanket claim is doing more work than it can support.

Nothing here is structural. I'd fix 1 and 2 before un-drafting, keep 7's site change and file the ownership issue, and treat 3 as a follow-up PR against `snapshot.rs`. 5 and 6 are dashboard polish.

brayniac (approved):
Re-read the branch at `35773ec4`. All seven items from my earlier review are settled — approving.

**1. `device` label** — Fixed. `devnum` (major:minor) plus `block_device` from `read_link("/sys/dev/block/<maj:min>")`, so nothing collides with `drivehealth`'s `device` any more.

**2. Silent resolution failure** — Fixed, and more thoroughly than I asked: `select_local()` returns a typed `SkipReason` per candidate plus `unresolvable_starts`, warned through `report_selection()` and gated on change. The `rezolus status` half is #1208.

**3. Per-metric `member_bound()`** — Deferred to #1207, which is the right home for it.

**4. Parked sweep** — Fixed. Latch checked before `last_read` is bumped, `warn_if_stuck()` at three intervals, and the first sweep now goes through `spawn_blocking` like every other. The `InFlight` drop-guard and the poisoned-mutex restart in `lock_state()` weren't in my list and are both improvements.

**5. Used %** — Fixed, `df`'s formula, pinned by a test.

**6. Inode divide-by-zero** — Fixed at the source, and I traced it rather than taking the description's word: `inode_gauge` → `i64::MIN` → `GaugeGroup::value` maps to `None` → the declared-group arm pushes `None` → null column → no sample on either side of the ratio. Correct end to end.

**7. Site version bump** — Out of the diff entirely, ownership tracked in #1209.

Both smaller notes handled: `MAX_MOUNTS` 256 with the warning gated on change, and the metrics.md I/O claim now says what btrfs actually does.

Two non-blocking leftovers, neither worth holding the merge:

- `docs/metrics.md:284-285` has an orphan table header — a bodyless `| Metric | Description | Metadata |` and separator sitting above the `block_device` paragraph, with the real header repeated at 291. Renders as an empty table. One-line delete whenever convenient.
- The join I originally wanted still isn't expressible: `/sys/dev/block/8:1` resolves to the partition (`sda1`), `drivehealth` publishes the drive (`sda`). The docs now say so plainly, which is the important part, but nothing tracks closing the gap. Fine to leave until someone asks for it.

Issue #1213, opened 2026-09-14
Collect PF-side hardware counters: the only view of link utilization under SR-IOV


brayniac:
## Summary

Rezolus has no source of NIC hardware (MAC/port) counters. Every traffic metric we
collect comes from netdev tracepoints, which means that under **SR-IOV there is no
vantage point in the fleet that can see physical link utilization**.

Split out of #1184, where it came up as an orthogonal gap: the double-count fix
neither causes nor worsens this, and the data source is different (ioctl/rtnetlink,
not BPF).

## Why netdev hooks cannot answer this

With a VF passed through to a guest, frames go guest VF driver -> the NIC's embedded
switch (VEB/eSwitch) -> wire. The hypervisor kernel never gets a `net_device` for that
traffic, so `net_dev_start_xmit` and `netif_receive_skb` never fire for it. Eight VFs
saturating a 25G port reads as near-zero on the host. That is not a filtering choice —
there is nothing to hook.

Two errors, in opposite directions:

- **Host is blind to VF traffic.** Only PF traffic is visible, which on a VF-heavy host
  is close to none of it.
- **Guests over-report the wire.** VF-to-VF on the same card is hairpinned by the
  embedded switch and never reaches the link, but each guest counts it as egress at its
  VF driver (which has a PCI parent, so it lands in `network_host_bytes` too). Summing
  guests overcounts by whatever was hairpinned.

No single agent sees the truth: the host can't see VFs, and no guest can see the others'.

## Where the number actually lives

- **Port-level MAC counters via ethtool `-S`** — e.g. mlx5 `rx_vport_*`/`tx_vport_*`,
  ixgbe `rx_pkts_nic`/`tx_pkts_nic`. These count everything the MAC moved, VFs included.
- **Per-VF stats over rtnetlink** — `IFLA_VF_STATS`, what `ip -s link show` prints. Gives
  attribution per VF rather than just a port total.

## What exists today

`network_ethtool` (`src/agent/samplers/network/linux/ethtool/`) already does ethtool
ioctls, but it is ENA-specific by construction: it looks up four AWS allowance counters
(`network_ena_{bandwidth,pps,conntrack,linklocal}_allowance_exceeded`) by stat name and
tracks nothing else. Generalizing it is plausibly the smallest path, but the stat names
are per-driver, so it needs a per-driver mapping table rather than a fixed list.

## Constraints

- Userspace reads, so **principle 17** applies — throttle and decouple from the scrape
[...]

Issue #1184, opened 2026-09-10
network_bytes/network_packets double-count TX on bonded and other stacked interfaces


brayniac:
## Summary

On a host with stacked netdevs — a bond being the common case, but also VLAN
interfaces, bridges, team, macvlan, and tunnels — `network_bytes` and
`network_packets` on the **transmit** side count the same packet once per
netdev it traverses. A bonded host reports roughly **2x** its real egress.

## Mechanism

`network_traffic` counts TX from the `net_dev_start_xmit` raw tracepoint
(`src/agent/samplers/network/linux/traffic/mod.bpf.c:50`), which the kernel
fires in `xmit_one()` — i.e. **once per `net_device` the skb is handed to**,
not once per packet leaving the host:

```
dev_hard_start_xmit(bond0) -> xmit_one() -> trace_net_dev_start_xmit(skb, bond0)  # counted
  bond_start_xmit -> bond_dev_queue_xmit -> __dev_queue_xmit(skb->dev = eth0)
    dev_hard_start_xmit(eth0) -> xmit_one() -> trace_net_dev_start_xmit(skb, eth0)  # counted again
```

The same shape applies to `vlan_dev_hard_start_xmit`, bridge forwarding,
macvlan, and any tunnel device — each layer is a real netdev with its own
`ndo_start_xmit`, so each layer trips the tracepoint. Stack them (VLAN over
bond) and it is 3x.

Because we sum over all CPUs into a single system-wide counter with no device
dimension, there is nothing downstream that can undo this.

## The RX side is (probably) not symmetric, which is worse

`trace_netif_receive_skb()` sits near the top of `__netif_receive_skb_core()`,
**above** the `another_round:` label. Bonding, bridging, and VLAN untagging all
re-enter via `goto another_round` inside that same call, so RX should be
counted once — with `skb->dev` still the physical slave.

If that's right, then on a bonded host `network_bytes{direction="receive"}` is
correct while `network_bytes{direction="transmit"}` is doubled, so the metric
is not merely inflated, it is *asymmetrically* inflated. Worth confirming on a
real bonded box before fixing, since the fix differs depending on the answer.

## Impact

- `crates/dashboard/src/dashboard/overview.rs:37` and `network.rs:14` —
  "Network transmit (bits/sec)" reads ~2x on any bonded or VLAN'd host.
- `crates/dashboard/src/dashboard/overview.rs:182` — transmit bitrate per unit
  of work is inflated by the same factor.
[...]

PR #1214, opened 2026-09-15
feat(network): count north/south traffic once, at the host boundary

brayniac:
## Summary

- Adds `network_host_bytes` / `network_host_packets` (`direction={receive,transmit}`) — the same traffic as `network_bytes`/`network_packets`, counted **once** at the netdev bound to a device driver instead of once per netdev traversed. Fixes the ~2x TX inflation on bonded hosts (3x for VLAN-over-bond) described in #1184.
- **Additive, not a filter.** Principle 9: summing is reversible, un-summing is not. A predicate in the probe discards packets permanently and makes every recording taken in the meantime unfixable. The sampler used 4 of its 8 `COUNTER_GROUP_WIDTH` slots, so both totals fit at **zero extra memory**.
- Opens no new cardinality — no per-interface keying, no schema churn, no ringbuf.

## Why a bus-parent test and not `priv_flags`

The obvious version is a trap. `IFF_BONDING` is set on the **physical slave** as well as the master:

```c
bond_main.c:6005  bond_dev->priv_flags  |= IFF_BONDING | ...   /* master */
bond_main.c:2135  slave_dev->priv_flags |= IFF_BONDING;        /* the NIC! */
```

Skipping on it doesn't halve a bonded host's TX, it **zeroes** it — a 2x overcount traded for a 0x undercount, silently. A priv_flags allowlist also misses tunnels (vxlan/gre/ipip set no distinguishing bit) and ages as stacking drivers land. `dev->dev.type->name` is no good as a predicate either: `net/wireless/core.c:1474` sets `wiphy_type`, so a wifi NIC would read as virtual.

`dev->dev.parent` — set by `SET_NETDEV_DEV()`, exactly what `/sys/class/net/<if>/device` reflects — has none of those problems, handles tunnels correctly (vxlan skipped; the encapsulated frame counted once on the real NIC), and gets bonding right for free since only the slave has a parent.

Verified against v6.12 by counting real `SET_NETDEV_DEV(` calls: **bond, VLAN, bridge, veth, tun/tap, macvlan, ipvlan, vxlan, dummy, loopback** set none; **`virtio_net` does** (`&vdev->dev`), so a VM guest counts its own egress rather than reporting zero.

## "Host boundary", not "wire"

Deliberate. On bare metal they coincide. In a guest the boundary is the hypervisor — virtio egress may be hairpinned to another VM on the same box without touching a link, and the guest cannot tell. The metric claims only what this vantage point can support, and the description says so.

Scope limits, all stated in `docs/metrics.md`:

- **Container east-west is absent.** veth → bridge → veth touches no bus-parented device. It appears in `network_bytes` (several times over) and not here.
- **VM east-west is not a gap in the same way** — each guest counts its own side at `virtio_net`; the hypervisor correctly reports zero because nothing left the box.
- **SR-IOV VF traffic is invisible to the hypervisor** — no netdev exists for it there, so no tracepoint fires, before or after this change. That needs PF-side hardware counters: #1213.

## Test plan

- [x] BPF compiles for **both** vendored architectures (Homebrew LLVM, `-target bpf`): x86_64 742,968 B, aarch64 660,656 B
- [x] Disassembly confirms the intended shape and cost — `netif_receive_skb`: 3 `probe_read_kernel` / 4 atomic adds; `net_dev_start_xmit`: 2 / 4. Hot-path cost is **one** extra probe-read on TX (`dev` is already an argument) and **two** on RX (no `dev` argument, so it chases `skb->dev` first)
- [x] Both arches' `vmlinux.h` carry `net_device.dev` and `device.parent`
- [x] `cargo test` 702 unit + 17 integration; clippy and fmt clean
- [x] `network_host_*` added to `METRIC_SAMPLERS` — not a prefix of any sampler name, so `extract-features` would otherwise file it as `unattributed`

**Not verified, deliberately flagged rather than implied:**

- [ ] **The BPF verifier has not run.** Compiling is not loading; that needs a real kernel. CI's `build-ubuntu-*` jobs compile the program and typecheck `mod.rs` (which has never been typechecked — it's Linux-only and I developed this on macOS), but nothing here loads it.
- [ ] **No overhead measurement**, which principle 16 asks for on a sampler change. The instruction-shape delta above is an argument, not a number.
- [ ] **No runtime check that a bonded host's 2x actually becomes 1x.** The mechanism is reasoned from kernel source, per the call on #1184 that code reading was sufficient.

Refs #1184.

https://claude.ai/code/session_01Ed86TVPGnQAxyFmgrdJpqi


brayniac:
## The three unchecked boxes, now resolved on hardware

This PR's test plan listed three things as deliberately unverified. Ran current `main` on a
bonded host (Debian 13, kernel 6.12.90, zen4) after merge:

- [x] **BPF verifier** — loads clean. 30 samplers, 23 active, **0 failed**, no verifier/libbpf
      errors in the agent log. CI compiles the program but never loads it; it is now loaded.
- [x] **Bonded 2x becomes 1x** — measured **3.04x** inflation (that host is br0 → bond0 → slave,
      a three-layer stack), and `network_host_bytes` matched the NIC drivers' own
      `tx_bytes` at **1.000**.
- [x] **Overhead** — measured, and it is a regression: **+28.8 ns/run (+64%)**, 44.8 -> 73.6
      ns/run against this PR's parent. Tracked in **#1218** with methodology and a proposed
      `tp_btf`/`fentry` fix.

Full numbers, including the ambient control window that showed `network_bytes` running ~1000x
`network_host_bytes` at idle on a VM host, are in
https://github.com/iopsystems/rezolus/issues/1184#issuecomment-5675580553.

That ambient result is worth flagging back here specifically: it is direct evidence for the
additive design this PR argued for on principle 9 grounds. The two counters differ by three
orders of magnitude on a hypervisor, so they are two different quantities rather than a broken
one and a fixed one — a filter would have destroyed the dominant term.

PR #1219, opened 2026-09-15
feat(viewer): surface host-boundary traffic above the per-interface pair

brayniac:
## Summary

#1214 added `network_host_bytes`/`network_host_packets` and **nothing displayed them**, so every dashboard still read the per-interface sum. On hardware that is measurably wrong: **3.04x** on a bonded host, and **~1000x** at idle on one running VMs, where loopback and virtual-interface traffic dominate ([measurements](https://github.com/iopsystems/rezolus/issues/1184#issuecomment-5675580553)).

Both dashboards and the TUI now plot the host-boundary series **above** the older panels, rather than replacing them.

## Why additive here too

Same reason the counters were. The two measure genuinely different things, and on a container or VM host the per-interface pair is the *only* thing that sees intra-host traffic at all — replacing it would have made pod-to-pod and VM-to-VM traffic vanish from the UI entirely.

To make the distinction legible without reading a description, the old groups are renamed **"Interface Bandwidth"** / **"Interface Packets"**, and each carries a note on why it reads higher.

One exception: the overview's bits-per-unit-of-work ratios **switch outright** instead of gaining a second copy. Wire efficiency is a wire question, and a duplicated ratio would be noise.

## The gate, which is the part worth reviewing

A query for an absent metric renders an *empty plot*, so a naive change would put two blank panels at the top of every recording made before 5.20.1. New `has_host_traffic_counters` in `plot.rs`, next to the existing `blockio_device_latency_metric` — same single-source-of-truth pattern, because three consumers need this decision and must not disagree about it.

Checked on `network_host_bytes` alone: all four counters come from one sampler reading one BPF map, so a recording has all of them or none.

## Descriptions

Amends `network_bytes`/`network_packets`' own descriptions to say they count once per interface traversed, name the 2x/3x cases, and point at `network_host_bytes`. Descriptions travel *into recordings*, so this reaches anyone reading a capture in the viewer, MCP, or `describe-metrics`.

Worth knowing: only for recordings made from here on. Existing recordings carry the old text — the metric-descriptions map is embedded at record time, not resolved at read time.

## Test plan

- [x] `cargo test` 702 + 17 integration; `cargo test -p dashboard` 91; `cargo test -p viewer` 14; clippy and fmt clean
- [x] `cargo check -p rez --no-default-features --target wasm32-unknown-unknown` (the WASM CI step) passes
- [x] New tests pin both halves of the gate: `host_panels_are_absent_when_the_recording_predates_them` (and asserts the interface pair is still plotted, so the view isn't empty) and `host_panels_are_present_and_come_before_the_interface_panels` — the latter asserts on *byte offsets in the rendered JSON*, so it fails if the ordering is ever flipped back
- [x] `crates/dashboard` is shared by the server and WASM viewers, so parity holds by construction — one change covers both. The TUI is a separate site and is updated too

**Not verified:** no visual check of the rendered dashboards. The tests assert query text and ordering in the JSON, not pixels — `viewer-render` would be the tool for that if the layout of two extra subgroups is a concern.

Refs #1184.

https://claude.ai/code/session_01Ed86TVPGnQAxyFmgrdJpqi


brayniac:
Filed **#1220** for the gap this PR can't close.

Amending a description only helps recordings made *after* it lands — descriptions are embedded at
record time, not resolved at read time — so the captures that most need the warning are precisely
the ones that cannot receive it. A recording from a pre-5.20.1 agent still says *"The number of
bytes transferred over the network"* over a 3x chart, with no `network_host_bytes` beside it to
compare against.

#1220 proposes a read-side, version-keyed errata table for that, and notes the enabler: #1211 put
the agent version into recording metadata this cycle, so a reader can finally tell which errata
apply to the file in front of it.

Issue #1203, opened 2026-09-12
Tegra: only the GPU's temperature and power rail are collected; CPU/SoC thermal zones, input power, fan and NIC temperature are not


thinkingfish:
### Summary

On NVIDIA Tegra hosts (Jetson Orin, Thor), the `gpu` sampler collects the GPU die temperature and the GPU power rail through NVML, as the Thor recording behind #1105 and #1108 shows. The rest of the board's thermal and power picture is not collected: CPU and SoC thermal zones, the system input power rail, fan speed, and the on-board NIC temperature. A recording from a Tegra host can therefore show a GPU that is cool and lightly loaded while the CPU cluster is thermally throttling or the fan has stopped.

### What the platform exposes

| signal | where the platform reports it |
| --- | --- |
| CPU cluster temperature | `tegrastats` `cpu@…C`; thermal zones under `/sys/class/thermal` |
| SoC temperature | `tegrastats` `soc0@…C`; thermal zones |
| GPU die temperature | `tegrastats` `soc2@…C`; already collected by Rezolus via NVML as `gpu_temperature` |
| system input power | `tegrastats` `VIN_SYS_5V0 … mW`; INA3221 rails under hwmon |
| CPU and GPU/SoC rails | `tegrastats` `VDD_CPU_CV`, `VDD_GPU_SOC`; the GPU rail is already `gpu_power_usage` |
| fan speed | hwmon `pwmfan` device, `pwm1` on a 0 to 255 scale |
| NIC temperature | hwmon device for the on-board Ethernet controller, `temp1_input` |

`tegrastats` requires root and streams continuously; the hwmon and thermal-zone nodes are plain sysfs reads.

### Expected behavior

On a Tegra host, Rezolus reports temperatures for the CPU and SoC zones alongside the GPU, board input power alongside the GPU rail, and fan speed, so the thermal and power state of the whole SoC is in the recording rather than only the GPU's share of it.

### Actual behavior

Only `gpu_temperature` and `gpu_power_usage` are collected. CPU and SoC temperatures, board input power, fan speed and NIC temperature have no Rezolus source.

### Environment

- NVIDIA Jetson, JetPack 6.1 (L4T r36.4), aarch64, kernel with BTF enabled.
- The same GPU-only coverage is visible in the Thor recording discussed in #1105.

_Generated by Claude on Yao's behalf._

PR #1256, opened 2026-09-19
feat(agent): add Linux hardware sensors with Tegra rail decoding

thinkingfish:
## Problem

Rezolus exposes GPU temperature and power through NVML, but lacks the CPU/SoC temperatures, board power rails, and fan readings requested in #1203. Thor's rail layout also cannot be assumed to apply to Orin.

## Solution

Add an opt-in Linux `sensors` sampler for native thermal/hwmon readings, with INA3221, INA238, and NVIDIA tachometer decoding. Preserve native channel identity and board context; apply system-power interpretation only to the verified Thor layout. Collect temperature, power, voltage, current, fan RPM, PWM, and cooling state, with a dashboard chart for each family.

Device reads run in one blocking worker at a configurable interval (default 5 seconds), with 60-second rediscovery. Slots retain their identity for the process lifetime. Include the supplied Thor inventory fixture, synthetic Orin tests, a read-only inventory script, and hardware-validation instructions.

## Result

Addresses the sensor-collection portion of #1203. Kept in draft pending Thor/Orin hardware validation and representative overhead measurements. The sampler remains disabled by default. Local x86 device sweeps measured 20–43 ms off the scrape path; these do not establish Tegra performance.

If every channel in a family fails, live readings clear but its acquisition window stays unchanged. Recorders that deduplicate windows can retain the prior observation without an outage marker. Configuration histories such as thermal thresholds and fan modes are outside this version's recorded metrics.

## Validation

Rebased onto `main` at `2875834a`; all four feature commits applied unchanged, without conflicts.

- Agent binary tests: 869 passed, 8 ignored, zero failures.
- Dashboard tests: 96 passed, zero failures.
- Locked agent build, formatting, and diff checks passed.
- Before the rebase: independent local review, an 80-second live sampling run through rediscovery, and a final-binary smoke test.
- Thor fixture and synthetic Orin coverage do not replace target-hardware testing; see `docs/sensors-validation.md` and `docs/journal/2026-09-17-sensors.md`.


src/agent/samplers/hw_sensors/linux/discovery.rs
```diff
@@ -0,0 +1,645 @@
...
+    let board = Board::read(sysfs);
+    discover_thermal(sysfs, &mut found);
+    discover_cooling(sysfs, &mut found);
+    discover_hwmon(sysfs, &board, &mut found);
+    for descriptor in &mut found.descriptors {
+        descriptor.board_model.clone_from(&board.model);
+        descriptor.soc_compatible.clone_from(&board.compatible);
+        // These are recorded series context. Include them in the published
+        // identity too, so a metadata replacement can never relabel a slot.
+        if let Some(model) = &board.model {
```
brayniac: Appending `:board_model=...:soc_compatible=...` to every descriptor's `sensor` makes the only label the dashboard groups by (`sum by (sensor) (...)`) mostly host-constant filler.

For the Thor fixture the INA238 system-power key becomes:

```
hwmon:/devices/platform/bus@0/c600000.i2c/i2c-2/2-0044:ina238:power1:scope=system:board_model=NVIDIA Jetson AGX Thor Developer Kit:soc_compatible=nvidia,p4071-0000+p3834-0008,nvidia,p3834-0008,nvidia,tegra264
```

211 characters, of which the last ~100 are byte-identical across every sensor on the host. All seven charts render legends where the distinguishing part (`ina238:power1`) sits mid-string.

The comment's reason — "so a metadata replacement can never relabel a slot" — doesn't hold: `/sys/firmware/devicetree/base/{model,compatible}` can't change within a process, so the suffix adds no identity discrimination. Both values are already published separately as `board_model` / `soc_compatible` metadata.

src/agent/samplers/hw_sensors/linux/discovery.rs
```diff
@@ -0,0 +1,645 @@
...
+        derived,
+        input,
+    });
+}
+
+fn device_identity(sysfs: &Path, node: &Path, hwmon: bool) -> String {
+    let canonical = std::fs::canonicalize(node).unwrap_or_else(|_| node.to_path_buf());
+    let in_class = canonical.starts_with(sysfs.join("class"));
+    let mut device = canonical.as_path();
+    if hwmon
```
brayniac, line 603: This strips the `hwmonN` component and then a parent literally named `hwmon`. For a driver that registers with a NULL parent the canonical path is `/sys/devices/virtual/hwmon/hwmonN`, which collapses to `/devices/virtual` for *every* such chip.

Two virtual hwmon instances with the same `name` and the same channel then produce an identical `sensor` string. `Slots::reconcile` finds the first in `by_sensor` and overwrites `descriptors[slot]`, so one of the two devices is dropped silently — no overflow warning, no error.

Guarding with the class-relative name, or keeping the `hwmonN` index when the parent chain yields no distinguishing component, would close it.

src/agent/samplers/hw_sensors/linux/discovery.rs
```diff
@@ -0,0 +1,645 @@
...
+                    &channel,
+                    label,
+                    scope,
+                    false,
+                    Input::Direct(input),
+                );
+            }
+        }
+
+        for name in attribute_names(&node, &mut found.errors) {
```
brayniac: `attribute_names(&node, &mut found.errors)` runs a second time on a node already scanned into `names` at line 285. Besides the redundant `read_dir`, a node whose directory can't be read pushes its error into `found.errors` twice, so `report_discovery` logs `discovery had 2 metadata/permission error(s)` for a single failure. The two passes can also observe different directory contents.

scripts/sensor_inventory.py
```diff
@@ -0,0 +1,59 @@
...
+import platform
+from datetime import datetime, timezone
+from pathlib import Path
+
+
+def read(path):
+    try:
+        return Path(path).read_text().replace("\0", "\n").strip()
+    except OSError as error:
+        return {"error": str(error)}
```
brayniac: Returning `{"error": str(error)}` on `OSError` means that on any host without a device tree (every x86 host, and Jetson under a container that hides `/sys/firmware`) the emitted JSON has object-valued `"model"` / `"compatible"`.

`docs/sensors-validation.md` documents this script as the way to capture a new fixture, but `Inventory` in `src/agent/samplers/hw_sensors/linux/tests.rs:14-15` declares both as `String`, so `serde_json::from_str` fails with `invalid type: map, expected a string`. Emitting `null` (or omitting the key) on error would let a captured inventory round-trip into a fixture.

src/agent/samplers/hw_sensors/linux/discovery.rs
```diff
@@ -0,0 +1,645 @@
...
+    family: Family,
+    source: &str,
+    chip: &str,
+    channel: &str,
+    label: Option<String>,
+    scope: Option<String>,
+    derived: bool,
+    input: Input,
+) {
+    let mut sensor = format!("{source}:{device}:{chip}:{channel}");
```
brayniac, line 572: The lifetime-slot cap (`MAX_SENSORS = 256` per family, slots never reused) is keyed on an identity that includes the resolved device path. NVMe hot-swap renumbers `/devices/.../nvme/nvmeN`, so each re-add of a drive mints a new temperature identity; a host that cycles drives exhausts the temperature family's 256 lifetime slots and then publishes nothing new for it.

The overflow warning fires, so this is visible rather than silent, but there's no recovery short of restarting the agent.

brayniac (commented):
Review notes on the hw_sensors sampler. None of these block the merge — the hwmon decoding, the acquisition-group ordering, the unit scalings and the `Slots` lifetime logic all check out. The label-shape one is worth doing before this lands because it changes what the published series key looks like; the rest are follow-ups.

What I checked and found correct, for the record:

- INA3221 `in1..3` = bus voltage (mV), `in4..7` shunt/sum, `curr4` summation; INA238 `in0` shunt / `in1` bus; derived power `mV x mA = uW`. The dashboard's `/1000`, `/1000000` and `/255` scalings match the declared storage units, and `Unit::Power` expects watts, which is what the query yields.
- The blocking sweep is the sole window writer, and ordering is metadata -> value -> `set_member_bound` -> `finish()`, which is the correct stamp-last order. Group names satisfy the `<sampler>_<shortname>` rule that `group_registry()` `debug_assert!`s, and `bpf_sampler_name` resolution matches `attribute_sampler`'s module-path resolution on Linux and non-Linux.
- `i64::MIN` is metriken's absent sentinel for `GaugeGroup` and `insert_metadata` survives it, so removed/failed slots publish as absent while keeping their labels.
- `Slots` lifetime-slot logic (`reconcile` / `pending_removed` / `metadata_applied`) is consistent in every ordering I could construct, and `refresh()` is driven serially from `SnapshotBuilder::refresh`, so the latch/throttle pair has no race.
- The `extract` golden was updated in lockstep with `EXPECTED_SUBSYSTEMS`, and `DOMAIN_ALIASES`'s `("sensor", "hw")` makes `domain_of("sensor_temperature") == domain_of("hw_sensors")`.

brayniac (approved):
Inline feedback is non-blocking. Approved.

thinkingfish:
Addressed the review feedback in f9793877:

- Removed `board_model` and `soc_compatible` from the `sensor` identity while retaining both as separate metadata, shortening chart legends.
- Kept the `hwmonN` path component for parentless virtual hwmon devices so otherwise identical devices receive distinct slots.
- Reused the discovered attribute list for PWM channels, avoiding a second directory scan and duplicate discovery errors.
- Changed unavailable inventory values to `null`, with diagnostics preserved separately in `read_errors`. Fixture loading now accepts missing platform metadata and skips unavailable attributes.

Added regression coverage for the identity collision, board context, and inventory handling. Validation: 872 Rust tests passed, 8 ignored; both Python tests passed; Clippy with warnings denied, formatting, and diff checks passed. Also checked a fresh inventory on an x86 host without device-tree metadata. Thor/Orin hardware validation remains pending.

The lifetime-slot exhaustion finding remains a follow-up. Documented the 256-identities-per-family limit, device-renumbering behavior, overflow warning, and restart recovery in `docs/sensors-validation.md`. Slots remain unreused to preserve recorded sample identities; safe capacity recovery needs a separate design.

